PR #11, opened 2019-11-29
Add App Store Connect API client

priitlatt:
Add [App Store Connect API](https://developer.apple.com/documentation/appstoreconnectapi) client and implement the endpoints under the Provisioning topic.

Minor changes were introduced to the `grab` tool, but this is still a work in progress and will be carried over to next PR.


src/codemagic_cli_tools/apple/app_store_connect/api_session.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .api_error import AppStoreConnectApiError
+
+if TYPE_CHECKING:
+    from codemagic_cli_tools.apple import AppStoreConnectApiClient
+
+
+class AppStoreConnectApiSession(requests.Session):
+
+    def __init__(self, app_store_connect_api: AppStoreConnectApiClient):
```
madispp: IMO it would be better to avoid circular dependency like this, at least for the whole class. Would it be possible to pass the headers to the session more explicitly, e.q. by passing a factory function or object?

src/codemagic_cli_tools/apple/resources/resource.py
```diff
@@ -0,0 +1,186 @@
...
+    BUNDLE_ID_CAPABILITIES = 'bundleIdCapabilities'
+    CERTIFICATES = 'certificates'
+    DEVICES = 'devices'
+    PROFILES = 'profiles'
+
+
+@dataclass
+class PagingInformation(DictSerializable):
+    @dataclass
+    class Paging(DictSerializable):
```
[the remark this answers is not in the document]
priitlatt: This is indeed to follow the API response structure. Check the [PagingInformation](https://developer.apple.com/documentation/appstoreconnectapi/paginginformation) reference, it too has a single `paging` property.

src/codemagic_cli_tools/apple/app_store_connect/api_session.py
priitlatt: Resolved in 509d040.

PR #21, opened 2020-02-10
Bugfix: Fix Xcode log copying

priitlatt:
Capturing unprocessed `xcodebuild` log into log file was not working properly and the tail of the actual log was not always included.

### DIFF
--- a/src/codemagic/models/xcodebuild.py
+++ b/src/codemagic/models/xcodebuild.py
@@ -55,8 +55,13 @@ class Xcodebuild:
         with self.logs_path.open('a') as fd, xcodebuild_cli_process.log_path.open('r') as process_logs:
             fd.write(f'>>> {xcodebuild_cli_process.safe_form}')
             fd.write('\n\n')
-            for chunk in process_logs.read(8192):
+
+            chunk = process_logs.read(8192)
+            while chunk:
                 fd.write(chunk)
+                chunk = process_logs.read(8192)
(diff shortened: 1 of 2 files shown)


src/codemagic/models/xcodebuild.py
```diff
@@ -55,8 +55,13 @@ def _log_process(self, xcodebuild_cli_process: Optional[XcodebuildCliProcess]):
         with self.logs_path.open('a') as fd, xcodebuild_cli_process.log_path.open('r') as process_logs:
             fd.write(f'>>> {xcodebuild_cli_process.safe_form}')
             fd.write('\n\n')
-            for chunk in process_logs.read(8192):
+
+            chunk = process_logs.read(8192)
+            while chunk:
                 fd.write(chunk)
+                chunk = process_logs.read(8192)
+            fd.write(process_logs.read())
```
madispp: Why is this necessary?

priitlatt: Possible race-condition when the process is still writing logs to the file. Happens sometimes if underlying xcodebuild process exits unexpectedly and buffers are not properly flushed.

madispp: I recommend adding a comment

```suggestion
            # do an extra read in case xcodebuild exits unexpectedly and did not flush last buffer
            fd.write(process_logs.read()) 
```

PR #31, opened 2020-04-15
Feature: add option to show tool version

priitlatt:
Add `--version` option to apps to reflect the current installed version. For example:

```bash
> cli-tools feature/show-version keychain --version
Keychain installed at /Users/priit/.virtualenvs/cli-tools-shZ38mB-/bin/keychain
keychain 0.1.7
```


codemagic.yaml
```diff
@@ -4,11 +4,12 @@ workflows:
...
         REPOSITORY: codemagic-ci-cd/cli-tools
     name: Generate binary distribution wheel
     scripts:
       - git tag -a v$(python3 setup.py --version) -m "Release v$(python3 setup.py --version)"
-      - python3 setup.py bdist_wheel
+      - python3 setup.py sdist bdist_wheel
```
madispp, line 12: Why build the source distribution if it's not sent to PyPI nor included in artifacts?

priitlatt: Excellent point. Forgot to add the artifacts entry for this.

priitlatt: I included tarball artifact in ad2d561.

PR #40, opened 2020-06-16
Feature/pypi release yaml

BondarenkoStas:
Added a workflow to upload the dists to PyPI.
Current version is already updated and works like a charm https://pypi.org/project/codemagic-cli-tools/
Also maybe just to merge this workflow with `release`?


codemagic.yaml
```diff
@@ -37,6 +39,9 @@ workflows:
         gsutil -h "Content-Type:application/x-wheel+zip" \
                -h "Cache-Control:no-cache,max-age=0" \
                cp -a public-read "$SRC_PATH" "$DST_URL"
+
```
priitlatt, line 42: Perhaps do it as a script step with either a comment or echo statement in the header for better understanding what is being done here?

PR #42, opened 2020-08-05
Feature: Add option to specify where intermediate xcarchive is stored for `xcode-project build-ios`

priitlatt:
**Changes**:
- Add option to define where the `xcarchive` package is saved during `build-ipa` execution. It now defaults to `build/ios/xcarchive` whereas it used to be in `$TMPDIR`.
- Always log out the exact `xcodebuild` commands that are being executed for transparency, even without verbose mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Version 0.2.4
 
 **Improvements**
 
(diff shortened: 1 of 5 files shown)


priitlatt:
Example screenshot of the usage and modified output.:
Screenshot 2020-08-05 at 12 14 07

PR #44, opened 2020-08-19
Support OpenSSH private key format for certificate private keys

priitlatt:
In our [documentation](https://docs.codemagic.io/yaml/distribution/#setting-up-code-signing-for-ios) we suggest to generate the private keys for certificates with 

```bash
ssh-keygen -t rsa -b 2048 -f ~/Desktop/codemagic_private_key -q -N ""
```

However, this does not generate the key in PEM format that is currently only supported format by our tooling. To overcome that, add capabilities to deal with OpenSSH private keys too.


madispp (approved):
LGTM

PR #49, opened 2020-09-09
Bugfix: Set iCloudContainerEnvironment export option for `xcode-project build-ipa` if need be

priitlatt:
`xcodebuild -help` states that
> iCloudContainerEnvironment : String
If the app is using CloudKit, this configures the "com.apple.developer.icloud-container-environment" entitlement. Available options vary depending on the type of provisioning profile used, but may include: Development and Production.

The export options property list that is generated by `xcode-project use-profiles` does not include `iCloudContainerEnvironment` key in it as it is impossible to know at that time what scheme, target, config is going to be archived. Consequently at that time we have no way of knowing what entitlements should be applied.

However, once we have an Xcode archive, we can check the used entitlements from the generated `.app` within the `.xcarchive`. Based on that information we can update the export options before exporting archive to ipa.


mikhail-tokarev (approved):
LGTM

PR #54, opened 2020-10-20
Feature: Add capabilities to run tests with xcode-project

priitlatt:
**Add support to run iOS tests with `xcode-project`**.

**New actions**:
- [`xcode-project clean`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/xcode-project-run-tests/docs/xcode-project/clean.md)
- [`xcode-project default-test-destination`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/xcode-project-run-tests/docs/xcode-project/default-test-destination.md)
- [`xcode-project test-destinations`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/xcode-project-run-tests/docs/xcode-project/test-destinations.md)
- [`xcode-project junit-test-results`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/xcode-project-run-tests/docs/xcode-project/junit-test-results.md)
- [`xcode-project run-tests`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/xcode-project-run-tests/docs/xcode-project/run-tests.md)
- [`xcode-project test-summary`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/xcode-project-run-tests/docs/xcode-project/test-summary.md)

**Global changes**:
- Simplify access to currently invoked CLI app. Instead of passing around the `CliApp` instance to subsequent function/method calls, add a `RunningCliAppMixin` that can be extended in case a class needs access to currently invoked CLI app.

**Example executions of new nactions**

- Clean Xcode project / workspace
```bash
> xcode-project clean --project banaan.xcodeproj
Clean banaan.xcodeproj
Execute "xcodebuild -project banaan.xcodeproj -scheme banaan clean"

▸ Clean Succeeded
Cleaned banaan.xcodeproj
```

- Show default test destination for currently active Xcode
```bash
> xcode-project default-test-destination
Show default test destination for Xcode 12.0.1 (12A7300)
iOS 14.0 iPhone SE (2nd generation)
```

- List available test destinations for currently active Xcode
```bash
> xcode-project test-destinations --runtime 'iOS 14.0' 'iOS 13.2' --name 'iPhone 11 Pro'                                                                                                                               List available test devices
Runtime: iOS 13.2
- iPhone 11 Pro
- iPhone 11 Pro Max
Runtime: iOS 14.0
- iPhone 11 Pro
- iPhone 11 Pro Max
```

- Convert Xcode test result bundle to JUnit XML
```bash
[...]

PR #67, opened 2021-02-09
appstore builds information (list builds, get latest app store and testflight build numbers)

BondarenkoStas:
List builds information from app store connect, as well as get the latest testflight and app store build number

### New actions:

- [app-store-connect list-builds](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/testflight_builds/docs/app-store-connect/list-builds.md)
- [app-store-connect get-latest-testflight-build-number](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/testflight_builds/docs/app-store-connect/get-latest-testflight-build-number.md)
- [app-store-connect get-latest-app-store-build-number](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/testflight_builds/docs/app-store-connect/get-latest-app-store-build-number.md)

### Global changes:

- Additionally handle datetime in a format containing timezone timedelta

### Example executions of new actions:

- list builds for all application

```bash

> app-store-connect list-builds
Found 259 Builds matching specified filters: *.
-- Build --
Id: b393b321-5f55-4141-8d6c-08ccfe457449
Type: builds
Expired: True
Icon asset token: {'templateUrl': 'https://is4-ssl.mzstatic.com/image/thumb/Purple/v4/11/be/8c/11be8c17-e7fe-3f9a-018e-92232e2959a0/Icon-76@2x.png.png/{w}x{h}bb.{f}', 'width': 152, 'height': 152}
Min os version: 8.0
Processing state: VALID
Version: 22
Uploaded date: 2019-02-04 05:56:50-08:00
Expiration date: 2019-05-05 05:56:50-07:00
-- Build --

```

- list not expired builds for a specific app for a specific pre-release version reverse ordered by build version

```bash
> app-store-connect list-builds --application-id '1453997552' --pre-release-version '1.0.0' --ordering 'VERSION' --reverse --not-expired 
Found 1 Build matching specified filters: app=1453997552, expired=False, preReleaseVersion.version=1.0.0.
-- Build --
Id: a671690d-edc8-46c7-b36c-24a72e2d915c
Type: builds
Expired: False
Icon asset token: ImageAsset(templateUrl='https://is2-ssl.mzstatic.com/image/thumb/Purple114/v4/80/d3/39/80d339cc-1ce8-0987-e3c0-25fc8de3275a/Icon-83.5@2x.png.png/{w}x{h}bb.{f}', height=167, width=167)
Min os version: 10.2
Processing state: VALID
Version: 67
Uses non exempt encryption: False
Uploaded date: 2021-02-02 04:00:19-08:00
Expiration date: 2021-05-03 04:00:19-07:00
```

- get a specific build
```bash
> app-store-connect list-builds --build-id 'a671690d-edc8-46c7-b36c-24a72e2d915c'
Found 1 Build matching specified filters: id=a671690d-edc8-46c7-b36c-24a72e2d915c.
-- Build --
Id: a671690d-edc8-46c7-b36c-24a72e2d915c
Type: builds
Expired: False
Icon asset token: ImageAsset(templateUrl='https://is2-ssl.mzstatic.com/image/thumb/Purple114/v4/80/d3/39/80d339cc-1ce8-0987-e3c0-25fc8de3275a/Icon-83.5@2x.png.png/{w}x{h}bb.{f}', height=167, width=167)
Min os version: 10.2
Processing state: VALID
Version: 67
Uses non exempt encryption: False
Uploaded date: 2021-02-02 04:00:19-08:00
Expiration date: 2021-05-03 04:00:19-07:00
```

- get latest testflight build number
```bash
> app-store-connect get-latest-testflight-build-number '1501865727'
Found 2 Pre Release Versions matching specified filters: app=1501865727.
2
```

- get latest testflight build number for a specific version
```bash
> app-store-connect get-latest-testflight-build-number '1501865727' --pre-release-version '1.0'
Found 1 Pre Release Version matching specified filters: app=1501865727, version=1.0.
1
```

- get latest testflight build number, wrong app id
```bash
> app-store-connect get-latest-testflight-build-number '1501865728'
Did not find any Pre Release Versions matching specified filters: app=1501865728.
```

- get latest testflight build number
```bash
> app-store-connect get-latest-app-store-build-number '1453997552'
Found 1 App Store Version matching specified filters: *.
13
```

- get latest app store build number for a specific version
```bash
> app-store-connect get-latest-testflight-build-number '1501865727' --pre-release-version '1.0'
Found 1 App Store Version matching specified filters: versionString=1.0.
13
```


src/codemagic/apple/resources/resource.py
```diff
@@ -72,7 +72,7 @@ def __post_init__(self):
 class Links(DictSerializable):
     _OMIT_IF_NONE_KEYS = ('related',)
 
-    def __init__(_self, self: str, related: Optional[str] = None):
+    def __init__(_self, self: Optional[str] = None, related: Optional[str] = None):
```
BondarenkoStas, line 75: The change is required because few relationships are not specified in API, and they don't have `self` field, e.g.

```
            "perfPowerMetrics":{
               "links":{
                  "related":"https://api.appstoreconnect.apple.com/v1/builds/807b1541-9601-48da-8169-9130f93d4b90/perfPowerMetrics"
               }
```

src/codemagic/apple/resources/resource.py
```diff
@@ -202,7 +202,10 @@ def from_iso_8601(cls, iso_8601_timestamp: str) -> datetime:
     def from_iso_8601(cls, iso_8601_timestamp: Optional[str]):
         if iso_8601_timestamp is None:
             return None
-        return datetime.strptime(iso_8601_timestamp, '%Y-%m-%dT%H:%M:%S.%f%z')
+        try:
+            return datetime.strptime(iso_8601_timestamp, '%Y-%m-%dT%H:%M:%S.%f%z')
+        except ValueError:
+            return datetime.strptime(iso_8601_timestamp, '%Y-%m-%dT%H:%M:%S%z')
```
BondarenkoStas, line 208: iso 8601 without milliseconds in /builds api

```
            ...
            "uploadedDate":"2021-01-28T06:01:32-08:00",
            "expirationDate":"2021-04-28T06:01:32-07:00",
            ...
```

src/codemagic/apple/resources/enums.py
```diff
@@ -10,6 +10,18 @@ def __str__(self):
         return str(self.value)
 
 
+class BuildOrdering(_ResourceEnum):
```
BondarenkoStas, line 13: Basically needed to translate cli argument values e.g. PRE_RELEASE_VERSION to order api values e.g. preReleaseVersion

src/codemagic/apple/resources/build.py
```diff
@@ -0,0 +1,60 @@
...
+        betaAppReviewSubmission: Relationship
+        appStoreVersion: Relationship
+        icons: Relationship
+
+        betaGroups: Relationship
+        perfPowerMetrics: Relationship
+        diagnosticSignatures: Relationship
+
+@dataclass
+class ImageAsset(DictSerializable):
```
BondarenkoStas, line 53: keeping it to not list it as a separate resource, it's only used in this class

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,71 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        kwargs = {'ordering': ordering, 'reverse': reverse}
+        return self._list_resources(builds_filter, self.api_client.builds, should_print, **kwargs)
+
+    @cli.action('get-latest-testflight-build-number',
+        BuildArgument.APPLICATION_ID,
+        BuildArgument.PRE_RELEASE_VERSION)
+    def get_latest_testflight_build_number(self,
+                                           application_id: Union[str, ResourceId],
+                                           pre_release_version: Optional[str] = None,
+                                           should_print: bool = False) -> None:
```
BondarenkoStas, line 203: should print False so that only build number is echoed to use as
```
export BUILD_NUMBER=$(app_store_connect get-latest-testflight-build-number '1501865727')
```

return to keep python api

doc.py
```diff
@@ -240,7 +240,7 @@ def write_action_command_usage(self, documentation_generator: ToolDocumentationG
         self._write_command_usage(CommandUsageGenerator(documentation_generator).get_action_command_usage(action))
 
     def _write_command_usage(self, lines):
-        self.file.new_header(level=3, title='Usage')
+        self.file.new_header(level=3, title='Usage', add_table_of_contents='n')
```
BondarenkoStas, line 243: needed because otherwise it breaks (no header with previous level (2) to generate content table)
https://github.com/didix21/mdutils/issues/48#issuecomment-738326106

src/codemagic/apple/resources/resource.py
```diff
@@ -218,7 +224,9 @@ def to_iso_8601(cls, dt: datetime) -> str:
     def to_iso_8601(cls, dt: Optional[datetime]):
         if dt is None:
             return None
-        return dt.strftime('%Y-%m-%dT%H:%M:%S.%f')[:-3] + '+0000'
+        if dt.tzinfo == timezone.utc:
+            return dt.strftime('%Y-%m-%dT%H:%M:%S.%f')[:-3] + '+0000'
+        return dt.isoformat()
```
BondarenkoStas, line 229: apparently new format, build api returns `'2021-01-28T06:01:32-08:00'`, so change the output so that Build -> dict works fine

priitlatt:
Add description to pull request what was changed, what new actions were introduced with sample usage and output. See this [PR](https://github.com/codemagic-ci-cd/cli-tools/pull/54) for example.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                BuildArgument.REVERSE)
+    def list_testflight_builds(self,
+                              application_id: Optional[Union[str, ResourceId]] = None,
+                              expired: Optional[bool] = None,
+                              not_expired: Optional[bool] = None,
+                              build_id: Optional[Union[str, ResourceId]] = None,
+                              pre_release_version: Optional[str] = None,
+                              processing_state: Optional[BuildProcessingState] = None,
+                              build_number: Optional[str] = None,
+                              ordering: Optional[BuildOrdering] = None,
```
priitlatt, line 175: As stated elsewhere, I'm not convinced that we need ordering and if we do, then it should be implemented more generally for all `list_*` tasks.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                              reverse: Optional[bool] = False,
+                              should_print: bool = True) -> List[Build]:
+        """
+        List Testflight builds
+        """
+        try:
+            expired_value = get_binary_arguments_value(expired, not_expired)
+        except:
```
priitlatt, line 183: Plain `except:` clause is considered to be anti-pattern and instead you should catch as specific exception as possible.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+            raise BuildArgument.NOT_EXPIRED.raise_argument_error('Using mutually exclusive flags "--expired" and "--not-expired"')
+
+        builds_filter = self.api_client.builds.Filter(
+            app=application_id,
+            expired=expired_value,
+            id=build_id,
+            processing_state=processing_state,
+            version=build_number,
```
priitlatt, line 191: Is there any particular reason not to map CLI params directly to Apple's API filters. Here `version` vs `build_number` introduces somewhat of discrepancy.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        BuildArgument.APPLICATION_ID,
+        BuildArgument.PRE_RELEASE_VERSION)
+    def get_latest_testflight_build_number(self,
+                                           application_id: Union[str, ResourceId],
+                                           pre_release_version: Optional[str] = None,
+                                           should_print: bool = False) -> int:
+        """
+        Get latest Testflight build number for the given application
+        """
+        builds_filter = self.api_client.builds.Filter(app=application_id, pre_release_version__dot__version=pre_release_version)
```
priitlatt, line 207:
```suggestion
        builds_filter = self.api_client.builds.Filter(
            app=application_id,
            pre_release_version__dot__version=pre_release_version,
        )
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        builds = self._list_resources(builds_filter, self.api_client.builds, should_print)
+        latest_build_number = max([int(build.attributes.version) for build in builds])
```
priitlatt: In case of no builds, this will fail unexpectedly, doesn't it?

priitlatt:
```suggestion
        latest_build_number = max((build.attributes.version for build in builds), key=int)
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        self.echo(str(latest_build_number))
+        return latest_build_number
+
+    @cli.action('get-testflight-build',
+        BuildArgument.BUILD_ID)
```
priitlatt, line 218:
```suggestion
    @cli.action('get-testflight-build', BuildArgument.BUILD_ID)
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+            processing_state=processing_state,
+            version=build_number,
+            pre_release_version__dot__version=pre_release_version,
+        )
+        kwargs = {'ordering': ordering, 'reverse': reverse}
+        return self._list_resources(builds_filter, self.api_client.builds, should_print, **kwargs)
+
+    @cli.action('get-latest-testflight-build-number',
+        BuildArgument.APPLICATION_ID,
+        BuildArgument.PRE_RELEASE_VERSION)
```
priitlatt, line 199:
```suggestion
    @cli.action('get-latest-testflight-build-number',
                BuildArgument.APPLICATION_ID,
                BuildArgument.PRE_RELEASE_VERSION)
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                              application_id: Optional[Union[str, ResourceId]] = None,
+                              expired: Optional[bool] = None,
+                              not_expired: Optional[bool] = None,
+                              build_id: Optional[Union[str, ResourceId]] = None,
+                              pre_release_version: Optional[str] = None,
+                              processing_state: Optional[BuildProcessingState] = None,
+                              build_number: Optional[str] = None,
+                              ordering: Optional[BuildOrdering] = None,
+                              reverse: Optional[bool] = False,
+                              should_print: bool = True) -> List[Build]:
```
priitlatt, line 177:
```suggestion
    def list_testflight_builds(self,
                               application_id: Optional[Union[str, ResourceId]] = None,
                               expired: Optional[bool] = None,
                               not_expired: Optional[bool] = None,
                               build_id: Optional[Union[str, ResourceId]] = None,
                               pre_release_version: Optional[str] = None,
                               processing_state: Optional[BuildProcessingState] = None,
                               build_number: Optional[str] = None,
                               ordering: Optional[BuildOrdering] = None,
                               reverse: Optional[bool] = False,
                               should_print: bool = True) -> List[Build]:
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                BuildArgument.EXPIRED,
+                BuildArgument.NOT_EXPIRED,
+                BuildArgument.BUILD_ID_OPTIONAL,
+                BuildArgument.PRE_RELEASE_VERSION,
+                BuildArgument.PROCESSING_STATE,
+                BuildArgument.BUILD_NUMBER,
+                BuildArgument.ORDERING,
+                BuildArgument.REVERSE)
+    def list_testflight_builds(self,
+                              application_id: Optional[Union[str, ResourceId]] = None,
```
priitlatt, line 168: Type could be just `Optional[ResouceId]` if the argument is defined properly. The same applies to `build_id`.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +154,72 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        builds_filter = self.api_client.builds.Filter(app=application_id, pre_release_version__dot__version=pre_release_version)
+        builds = self._list_resources(builds_filter, self.api_client.builds, should_print)
+        latest_build_number = max([int(build.attributes.version) for build in builds])
+        self.echo(str(latest_build_number))
+        return latest_build_number
+
+    @cli.action('get-testflight-build',
+        BuildArgument.BUILD_ID)
+    def get_testflight_build(self,
+                             build_id: Union[str, ResourceId],
```
priitlatt, line 220:
```suggestion
                             build_id: Union[ResourceId],
```

src/codemagic/utilities/argument_utils.py
```diff
@@ -0,0 +1,8 @@
+def get_binary_arguments_value(value, not_value):
+    if value and not_value:
+        raise Exception
```
priitlatt, line 3: Too broad exception. `ValueError` with appropriate message would suit a lot more.

src/codemagic/utilities/argument_utils.py
```diff
@@ -0,0 +1,8 @@
+def get_binary_arguments_value(value, not_value):
```
priitlatt, line 1: This function does not work in expected way if both arguments are `False`. It should raise, but instead returns `None`.

priitlatt, line 1: Instead of creating new `argument_utils` module we could just add this functionality to `Argument` class:
```python
class Argument(ArgumentProperties, enum.Enum):

    @classmethod
    def resolve_optional_two_way_switch(cls,
                                        is_switched_on: Optional[bool],
                                        is_switched_off: Optional[bool]) -> Optional[bool]:
        if {is_switched_on, is_switched_off} in ({True}, {False}):
            raise ValueError('Neither of the switches, or exactly one can be truthy at the time')
        elif is_switched_on is True:
            return True
        elif is_switched_off is True:
            return False
        else:
            return None
```

tests/apple/app_store_connect/builds/test_builds.py
```diff
@@ -0,0 +1,35 @@
...
+    (BuildOrdering.VERSION, False, 'version'),
+    (BuildOrdering.UPLOADED_DATE, True, '-uploadedDate'),
+])
+def test_ordering(ordering, reverse, expected_ordering, api_client):
+    with mock.patch.object(api_client.builds.client, 'paginate', return_value=[]) as mock_paginate:
+        builds = api_client.builds.list(ordering=ordering, reverse=reverse)
+        mock_paginate.assert_called_once_with(mock.ANY, params={'sort': expected_ordering})
+
+
+@pytest.mark.skipif(not os.environ.get('RUN_LIVE_API_TESTS'), reason='Live App Store Connect API access')
```
priitlatt, line 22: Nice usage of `skipif` 👍

priitlatt (changes requested):

src/codemagic/apple/app_store_connect/builds/builds.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def list(self,
+             resource_filter: Filter = Filter(),
+             ordering: Optional[BuildOrdering] = None,
+             reverse: bool = False) -> List[Build]:
+        """
+        https://developer.apple.com/documentation/appstoreconnectapi/list_builds
+        """
+
+        params = resource_filter.as_query_params()
```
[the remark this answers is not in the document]
BondarenkoStas: i'd write in this way if ordering wasn't optional. But yes, some ordering won't hurt, will sort by `uploadedDate` by default and always pass `sort`

src/codemagic/apple/app_store_connect/builds/builds.py
```diff
@@ -0,0 +1,60 @@
...
+        return Build
+
+    @dataclass
+    class Filter(ResourceManager.Filter):
+        app: Optional[Union[str, ResourceId]] = None
+        expired: Optional[str] = None
+        id: Optional[Union[str, ResourceId]] = None
+        processing_state: Optional[BuildProcessingState] = None
+        version: Optional[str] = None
+        pre_release_version__dot__version: Optional[str] = None
```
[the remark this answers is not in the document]
BondarenkoStas: I like this approach!

src/codemagic/apple/resources/build.py
```diff
@@ -0,0 +1,59 @@
...
+        app: Relationship
+        appEncryptionDeclaration: Relationship
+        individualTesters: Relationship
+        preReleaseVersion: Relationship
+        betaBuildLocalizations: Relationship
+        buildBetaDetail: Relationship
+        betaAppReviewSubmission: Relationship
+        appStoreVersion: Relationship
+        icons: Relationship
+        betaGroups: Relationship
```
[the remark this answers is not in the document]
BondarenkoStas: they always come in build's response, it would already fail if they weren't present in at least one build.
Sure can make them optional with default `None` and redefine `asdict` function to not return `None` fields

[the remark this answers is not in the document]
priitlatt: Since they are not defined in API doc, I'd maybe define them like
```suggestion
        betaGroups: Optional[Relationship] = None
```
Of course we'd need to make sure that `asdict` and any other method would remain functional.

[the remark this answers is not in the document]
priitlatt: My main concern is that something along the lines of the following would happen:
```python
>>> @dataclass
... class MyDataclass:
...     i: int
...     s: str
>>> MyDataclass(i=1)
  File "<input>", line 5, in <module>
TypeError: __init__() missing 1 required positional argument: 's'
```

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -4,6 +4,9 @@
 from codemagic.apple.app_store_connect import AppStoreConnectApiClient
 from codemagic.apple.app_store_connect import IssuerId
 from codemagic.apple.app_store_connect import KeyIdentifier
+from codemagic.apple.app_store_connect.builds import Builds
```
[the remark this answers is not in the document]
BondarenkoStas: For some reason can't comment on previous ones

> Can't you reuse Builds.Ordering as CLI argument value insead?

I want to keep cli arguments consistently uppercase, while ordering is in camel

> I get the signature change, but instance attributes should be defined nonetheless. If attribute self is None, then it shouldn't just be added to dict. But in case the attributes are not defined entirely, this could have broad negative effect in many unforseen places.

right, it's better to not include `None` fields to `DictSerializable`'s `dict` method

> Could you elaborate what exactly is going on here?
```
In [9]: dt = datetime(2021, 2, 12)
In [10]: Resource.to_iso_8601(dt)
Out[10]: '2021-02-12T00:00:00'
In [11]: utc_dt = datetime(2021, 2, 12, tzinfo=timezone.utc)
In [12]: Resource.to_iso_8601(utc_dt)
Out[12]: '2021-02-12T00:00:00.000+0000'
```

So that both string date formats would be converted back to string from datetime to their original form
```python
>>> s_utc = '2020-08-04T11:44:12.000+0000'
>>> dt_utc = datetime.strptime(s_utc, '%Y-%m-%dT%H:%M:%S.%f%z')
>>> dt_utc
datetime.datetime(2020, 8, 4, 11, 44, 12, tzinfo=datetime.timezone.utc)
>>> dt_utc.strftime('%Y-%m-%dT%H:%M:%S.%f')[:-3] + '+0000'
'2020-08-04T11:44:12.000+0000'
>>>
>>> s_timedelta = '2021-01-28T06:01:32-08:00'
>>> dt_timedelta = datetime.strptime(s_timedelta, '%Y-%m-%dT%H:%M:%S%z')
>>> dt_timedelta
datetime.datetime(2021, 1, 28, 6, 1, 32, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=57600)))
>>> dt_timedelta.isoformat()
'2021-01-28T06:01:32-08:00'
```

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +99,91 @@ class AppStoreConnectArgument(cli.Argument):
     )
 
 
+class BuildArgument(cli.Argument):
+    APPLICATION_ID = cli.ArgumentProperties(
+        key='application_id',
+        description='Application Apple ID',
```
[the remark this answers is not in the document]
BondarenkoStas: yes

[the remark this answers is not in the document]
priitlatt: Could you be more elaborate on the description then?

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +99,91 @@ class AppStoreConnectArgument(cli.Argument):
...
+        type=bool,
+        description='List only not expired builds',
+        argparse_kwargs={
+            'required': False,
+            'action': 'store_true',
+        },
+    )
+    BUILD_ID = cli.ArgumentProperties(
+        key='build_id',
+        description='Specific build ID',
```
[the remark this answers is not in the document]
BondarenkoStas: Basically `builds` will return only one build which matches its resource id. Doesn't actually make sense to have it here while we have a separate action to get one build. But maybe it's better to remove the `get-testflight-build` action

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +99,91 @@ class AppStoreConnectArgument(cli.Argument):
...
+            'choices': list(BuildProcessingState),
+        },
+    )
+    BUILD_NUMBER = cli.ArgumentProperties(
+        key='build_number',
+        flags=('--build-number',),
+        description='Build number',
+        argparse_kwargs={'required': False},
+    )
+    ORDERING = cli.ArgumentProperties(
```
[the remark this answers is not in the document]
BondarenkoStas: don't know what's needed:) as for me, i don't know a real use case for this action because it doesn't seamlessly provide other resources like apps or versions. Deleting order won't hurt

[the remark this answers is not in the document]
priitlatt: I think just remove it for now from here and all related places. Later we can add more generic ordering support for all listing actions. It should make many things simpler.

src/codemagic/tools/app_store_connect.py
```diff
@@ -127,9 +133,9 @@ def _get_resource(self, resource_id, resource_manager, should_print):
         self.printer.print_resource(resource, should_print)
         return resource
 
-    def _list_resources(self, resource_filter, resource_manager, should_print):
+    def _list_resources(self, resource_filter, resource_manager, should_print, **kwargs):
```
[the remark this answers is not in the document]
BondarenkoStas: right, ordering will be deleted anyway. But instead of kwargs will put `limit` to be able to sort `builds` by `-uploadedDate` and will `limit: 1` get the build with latest build number as Fastlane does

src/codemagic/tools/app_store_connect.py
BondarenkoStas: apple uses `build version` for build number and `build appstore version/build prerelease version` for build version. Did that to not confuse users with `build_version` and `pre_release_version`

src/codemagic/utilities/argument_utils.py
BondarenkoStas: I made it for our case when argument can only be `True` or `None`, but yes, will do in a more general way

src/codemagic/apple/resources/resource.py
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-b736b487114b15d1320d4be383e9cb605e79ac0e683d6949d012fc6215dcf44eR74)

src/codemagic/apple/resources/resource.py
BondarenkoStas: added [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/files#diff-b736b487114b15d1320d4be383e9cb605e79ac0e683d6949d012fc6215dcf44eR204)

src/codemagic/apple/resources/enums.py
BondarenkoStas: no ordering, not relevant anymore:)

src/codemagic/apple/resources/build.py
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-51b8fb81410afec451e8bc989ee6a2cd03ab8b4f5df4f2423b4058c949bc2bd3R10)

src/codemagic/apple/resources/resource.py
BondarenkoStas: added an annotation [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-b736b487114b15d1320d4be383e9cb605e79ac0e683d6949d012fc6215dcf44eR227)

src/codemagic/apple/app_store_connect/builds/builds.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    @property
+    def resource_type(self) -> Type[Build]:
+        return Build
+
+    @dataclass
+    class Filter(ResourceManager.Filter):
+        app: Optional[Union[str, ResourceId]] = None
+        expired: Optional[str] = None
```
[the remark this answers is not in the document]
BondarenkoStas: right, fixed

src/codemagic/apple/app_store_connect/builds/builds.py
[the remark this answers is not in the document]
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-b736b487114b15d1320d4be383e9cb605e79ac0e683d6949d012fc6215dcf44eR227)

src/codemagic/apple/app_store_connect/builds/builds.py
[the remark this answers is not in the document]
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-65fde6e4d2496f4bfd5c30d3c9f5c4b3cb9a3d8aabf05be48e8b5c455ccd3346R52)

src/codemagic/apple/resources/build.py
```diff
@@ -0,0 +1,59 @@
...
+        iconAssetToken: ImageAsset
+        minOsVersion: str
+        processingState: BuildProcessingState
+        version: str
+        usesNonExemptEncryption: bool
+        uploadedDate: datetime
+        expirationDate: datetime
+
+        def __post_init__(self):
+            if isinstance(self.processingState, str):
```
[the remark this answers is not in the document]
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-51b8fb81410afec451e8bc989ee6a2cd03ab8b4f5df4f2423b4058c949bc2bd3R47)

src/codemagic/apple/resources/build.py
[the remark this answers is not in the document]
BondarenkoStas: fixed [here](7cf35a0d94ec2496fb6053f99fda415a08b96a6d)

src/codemagic/tools/_app_store_connect/arguments.py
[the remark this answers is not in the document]
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-caa1aae40f5faf0bf43c7eab968651cd6ce22681e07ed14d7ca18f2358fb781aR137)

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +99,91 @@ class AppStoreConnectArgument(cli.Argument):
...
+    BUILD_ID_OPTIONAL = cli.ArgumentProperties(
+        key='build_id',
+        flags=('--build-id',),
+        description='Specific build ID',
+        argparse_kwargs={'required': False},
+    )
+    PRE_RELEASE_VERSION= cli.ArgumentProperties(
+        key='pre_release_version',
+        flags=('--pre-release-version',),
+        description='Pre-release version of your application',
```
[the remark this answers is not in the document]
BondarenkoStas: nice [description](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-caa1aae40f5faf0bf43c7eab968651cd6ce22681e07ed14d7ca18f2358fb781aR144) :)

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +99,91 @@ class AppStoreConnectArgument(cli.Argument):
...
+        key='ordering',
+        flags=('--ordering',),
+        type=BuildOrdering,
+        description='Order builds by attribute',
+        argparse_kwargs={
+            'required': False,
+            'choices': list(BuildOrdering),
+        }
+    )
+    REVERSE = cli.ArgumentProperties(
```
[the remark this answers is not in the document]
BondarenkoStas: just removed

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +99,91 @@ class AppStoreConnectArgument(cli.Argument):
...
+        flags=('--not-expired',),
+        type=bool,
+        description='List only not expired builds',
+        argparse_kwargs={
+            'required': False,
+            'action': 'store_true',
+        },
+    )
+    BUILD_ID = cli.ArgumentProperties(
+        key='build_id',
```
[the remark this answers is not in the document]
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R161)

src/codemagic/tools/app_store_connect.py
BondarenkoStas: fix [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R178)

src/codemagic/tools/app_store_connect.py
BondarenkoStas: changed it to build_version_number [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-caa1aae40f5faf0bf43c7eab968651cd6ce22681e07ed14d7ca18f2358fb781aR163)

src/codemagic/utilities/argument_utils.py
BondarenkoStas: will just delete it

src/codemagic/utilities/argument_utils.py
BondarenkoStas: added [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-46d5d4f67d34f1a5bcd1808beeab3ac0dd2953aa339f92151b8b45b4553b2e31R18)

src/codemagic/tools/app_store_connect.py
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R166)

src/codemagic/tools/app_store_connect.py
BondarenkoStas: fixed [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/9179f38154ac3341df96330eb09402de47e4a368#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R207)

src/codemagic/tools/_app_store_connect/arguments.py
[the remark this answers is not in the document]
BondarenkoStas: I think [this](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/71f2ca10806aa46be85d0feca84182780939f3d7#diff-caa1aae40f5faf0bf43c7eab968651cd6ce22681e07ed14d7ca18f2358fb781aR111) should be fine

src/codemagic/apple/resources/resource.py
priitlatt: The comment would be better suited in `except` block.

src/codemagic/apple/app_store_connect/resource_manager.py
```diff
@@ -59,12 +63,16 @@ def __str__(self):
             restrictions = self._get_restrictions()
             if not restrictions:
                 return '*'
-            return ', '.join(f'{param}={shlex.quote(value)}' for param, value in restrictions.items())
+            return ', '.join(f'{param}={shlex.quote(str(value))}' for param, value in restrictions.items())
```
BondarenkoStas, line 66: need to ensure it's string since `expired` filter has boolean
```
>>> shlex.quote(True)
  File "<stdin>", line 1, in <module>
  File "/Users/stas/.pyenv/versions/3.8.0/lib/python3.8/shlex.py", line 324, in quote
    if _find_unsafe(s) is None:
TypeError: expected string or bytes-like object
>>> shlex.quote(str(True))
'True'
```

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,47 @@
...
+             include: Include = Include.BUILD) -> Tuple[List[AppStoreVersion], List[Build]]:
+        """
+        https://developer.apple.com/documentation/appstoreconnectapi/list_all_app_store_versions_for_an_app
+        """
+
+        params = {
+            'include': include.value,
+            **resource_filter.as_query_params(),
+        }
+        results = self.client.paginate_with_included(f'{self.client.API_URL}/apps/{application_id}/appStoreVersions', params=params)
```
priitlatt, line 43: Formatting:
```suggestion
        results = self.client.paginate_with_included(
            f'{self.client.API_URL}/apps/{application_id}/appStoreVersions', params=params)
```

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
```diff
@@ -0,0 +1,47 @@
...
+        return PreReleaseVersion
+
+    @dataclass
+    class Filter(ResourceManager.Filter):
+        app: Optional[ResourceId] = None
+        version: Optional[str] = None
+
+    @dataclass
+    class Include(ResourceManager.Include):
+        BUILDS = 'builds'
```
priitlatt, line 30:
```suggestion
    class Include(ResourceManager.Include):
        BUILDS = ('builds', Build)
```

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def list(self,
+             resource_filter: Filter = Filter(),
+             include: Include = Include.BUILDS) -> Tuple[List[PreReleaseVersion], List[Build]]:
```
priitlatt, line 34:
```suggestion
             include: Include = Include.BUILDS) -> Tuple[List[PreReleaseVersion], List[Resource]]:
```

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+
+        params = {
+            'include': include.value,
+            **resource_filter.as_query_params(),
+        }
+        results = self.client.paginate_with_included(f'{self.client.API_URL}/preReleaseVersions', params=params)
+        return (
+            [PreReleaseVersion(prerelease_version) for prerelease_version in results.data],
+            [Build(build) for build in results.included]
```
priitlatt, line 46:
```suggestion
            [include.resource_type(included) for included in results.included]
```

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,47 @@
...
+    def resource_type(self) -> Type[AppStoreVersion]:
+        return AppStoreVersion
+
+    @dataclass
+    class Filter(ResourceManager.Filter):
+        version_string: Optional[str] = None
+
+    @dataclass
+    class Include(ResourceManager.Include):
+        BUILD = 'build'
```
priitlatt, line 29:
```suggestion
    class Include(ResourceManager.Include):
        BUILD = ('build', Build)
```

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def list(self,
+             resource_filter: Filter = Filter(),
+             application_id: ResourceId = ResourceId(),
+             include: Include = Include.BUILD) -> Tuple[List[AppStoreVersion], List[Build]]:
```
priitlatt, line 34:
```suggestion
             include: Include = Include.BUILD) -> Tuple[List[AppStoreVersion], List[Resource]]:
```

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+
+        params = {
+            'include': include.value,
+            **resource_filter.as_query_params(),
+        }
+        results = self.client.paginate_with_included(f'{self.client.API_URL}/apps/{application_id}/appStoreVersions', params=params)
+        return (
+            [AppStoreVersion(app_store_version) for app_store_version in results.data],
+            [Build(build) for build in results.included]
```
priitlatt, line 46:
```suggestion
            [include.resource_type(included) for included in results.included]
```

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,47 @@
...
+    def list(self,
+             resource_filter: Filter = Filter(),
+             application_id: ResourceId = ResourceId(),
+             include: Include = Include.BUILD) -> Tuple[List[AppStoreVersion], List[Build]]:
+        """
+        https://developer.apple.com/documentation/appstoreconnectapi/list_all_app_store_versions_for_an_app
+        """
+
+        params = {
+            'include': include.value,
```
priitlatt, line 40:
```suggestion
            'include': include.include_name,
```

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def list(self,
+             resource_filter: Filter = Filter(),
+             include: Include = Include.BUILDS) -> Tuple[List[PreReleaseVersion], List[Build]]:
+        """
+        https://developer.apple.com/documentation/appstoreconnectapi/list_prerelease_versions
+        """
+
+        params = {
+            'include': include.value,
```
priitlatt, line 40:
```suggestion
            'include': include.include_name,
```

src/codemagic/apple/app_store_connect/resource_manager.py
```diff
@@ -60,12 +64,16 @@ def __str__(self):
...
+
 
     class Ordering(enum.Enum):
         def as_param(self, reverse=False) -> str:
             return f'{"-" if reverse else ""}{self.value}'
 
+    class Include(enum.Enum):
+        pass
```
priitlatt, line 75:
```suggestion
    class Include(Tuple[str, Type[Resource]], enum.Enum):
        @property
        def include_name(self) -> str:
            return self.value[0]

        @property
        def resource_type(self) -> Type[Resource]:
            return self.value[1]
```

priitlatt (changes requested):
Some intermediate suggestions.

src/codemagic/apple/resources/resource.py
```diff
@@ -202,7 +203,15 @@ def from_iso_8601(cls, iso_8601_timestamp: str) -> datetime:
...
             return None
-        return datetime.strptime(iso_8601_timestamp, '%Y-%m-%dT%H:%M:%S.%f%z')
+        try:
+            return datetime.strptime(iso_8601_timestamp, '%Y-%m-%dT%H:%M:%S.%f%z')
+        except ValueError:
+            """
+            while most of API responses contain timestamp as '2020-08-04T11:44:12.000+0000'
+            /builds endpoint returns timestamps with timedelta and without milliseconds
+            as '2021-01-28T06:01:32-08:00'
+            """
```
priitlatt, line 213:
```suggestion
            # while most of API responses contain timestamp as '2020-08-04T11:44:12.000+0000'
            # /builds endpoint returns timestamps with timedelta and without milliseconds
            # as '2021-01-28T06:01:32-08:00'
```

src/codemagic/apple/resources/resource.py
```diff
@@ -218,7 +227,16 @@ def to_iso_8601(cls, dt: datetime) -> str:
...
             return None
-        return dt.strftime('%Y-%m-%dT%H:%M:%S.%f')[:-3] + '+0000'
+        if dt.tzinfo == timezone.utc:
+            """
+            while most of API responses contain timestamps as '2020-08-04T11:44:12.000+0000'
+            resolved to datetime.datetime(2020, 8, 4, 11, 44, 12, tzinfo=datetime.timezone.utc),
+            /builds endpoint returns timestamps as isoformat() '2021-01-28T06:01:32-08:00'
+            resolved to datetime.datetime(2021, 1, 28, 6, 1, 32, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=57600))).
+            So need to convert it to the initial form based on the presense of the explicit utc timezone
+            """
```
priitlatt, line 237:
```suggestion
            # while most of API responses contain timestamps as '2020-08-04T11:44:12.000+0000'
            # resolved to datetime.datetime(2020, 8, 4, 11, 44, 12, tzinfo=datetime.timezone.utc),
            # /builds endpoint returns timestamps as isoformat() '2021-01-28T06:01:32-08:00'
            # resolved to datetime.datetime(2021, 1, 28, 6, 1, 32, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=57600))).
            # So need to convert it to the initial form based on the presence of the explicit utc timezone
```

tests/apple/app_store_connect/test_resource_manager.py
```diff
@@ -18,6 +18,7 @@ class StubFilter(ResourceManager.Filter):
     field_one: Optional[str] = None
     field_two: Optional[StubEnum] = None
     field_three_four: Optional[CustomString] = None
+    field_with__dot__dots: Optional[str] = None
```
priitlatt, line 21: This isn't relevant any more, is it?

priitlatt (changes requested):

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +153,65 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        """
+        builds_filter = self.api_client.builds.Filter(
+            app=application_id,
+            pre_release_version_version=pre_release_version,
+        )
+        builds = self._list_resources(builds_filter, self.api_client.builds, should_print)
+        try:
+            latest_build_number = max((build.attributes.version for build in builds), key=int)
+        except ValueError:
+            return None
```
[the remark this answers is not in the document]
BondarenkoStas: when there are no builds (empty list). In such case there will be log in cli e.g.
```
Did not find any Pre Release Versions matching specified filters: app=1501865728
```
pre release or app store version is always linked to some build, so the it only happens when there are no versions

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,47 @@
...
+    class Filter(ResourceManager.Filter):
+        version_string: Optional[str] = None
+
+    @dataclass
+    class Include(ResourceManager.Include):
+        BUILD = 'build'
+
+    def list(self,
+             resource_filter: Filter = Filter(),
+             application_id: ResourceId = ResourceId(),
```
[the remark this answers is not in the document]
BondarenkoStas: okay, but will require some more modifications since now it's just taken from kwargs

tests/apple/app_store_connect/test_resource_manager.py
BondarenkoStas: yes, removed https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/1e641150af41c98a49a55bacdfa22af935437d5a#diff-122a8c79fd95bacf0d08d430bcd827e432174b82ddb5ae6746b36287ed453970L21

src/codemagic/apple/resources/resource.py
BondarenkoStas: https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/1e641150af41c98a49a55bacdfa22af935437d5a#diff-b736b487114b15d1320d4be383e9cb605e79ac0e683d6949d012fc6215dcf44eR229

src/codemagic/apple/resources/resource.py
BondarenkoStas: https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/1e641150af41c98a49a55bacdfa22af935437d5a#diff-b736b487114b15d1320d4be383e9cb605e79ac0e683d6949d012fc6215dcf44eR209

src/codemagic/apple/app_store_connect/resource_manager.py
BondarenkoStas: https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/1e641150af41c98a49a55bacdfa22af935437d5a#diff-47fe117de0afac6624b4539ad2b5eb3a3b0cdfbec4cf9e46ed88338dc7177decR74

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
BondarenkoStas: https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/1e641150af41c98a49a55bacdfa22af935437d5a#diff-1d6b24cb98b9559accd7f6526c3be95867475f8d283833f4c4801af679707a92R40

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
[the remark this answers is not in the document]
BondarenkoStas: check in [this way](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/1e641150af41c98a49a55bacdfa22af935437d5a#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R138), removed kwargs

src/codemagic/apple/app_store_connect/resource_manager.py
```diff
@@ -66,6 +71,14 @@ class Ordering(enum.Enum):
         def as_param(self, reverse=False) -> str:
             return f'{"-" if reverse else ""}{self.value}'
 
+    class Include(Tuple[str, Type[Resource]], enum.Enum):
+        @property
+        def include_name(self) -> str:
+            return self.value[0]
+        @property
```
priitlatt, line 78:
```suggestion

        @property
```

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +97,89 @@ class AppStoreConnectArgument(cli.Argument):
...
+        argparse_kwargs={'required': False},
+    )
+    PRE_RELEASE_VERSION = cli.ArgumentProperties(
+        key='pre_release_version',
+        flags=('--pre-release-version',),
+        description=(
+            'Version of the build published to Testflight '
+            'that identifies an iteration of the bundle. '
+            'The string can only contain one to three groups of numeric characters (0-9) '
+            'separated by period in the format [Major].[Minor].[Patch]'
```
priitlatt, line 161: Maybe add an example too:
```suggestion
            'separated by period in the format [Major].[Minor].[Patch]. '
            'For example `3.2.46`'
```

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +97,89 @@ class AppStoreConnectArgument(cli.Argument):
...
 
+class AppStoreVersionArgument(cli.Argument):
+    APP_STORE_VERSION = cli.ArgumentProperties(
+        key='app_store_version',
+        flags=('--app-store-version',),
+        description=(
+            'Version of the build published to App Store '
+            'that identifies an iteration of the bundle. '
+            'The string can only contain one to three groups of numeric characters (0-9) '
+            'separated by period in the format [Major].[Minor].[Patch]'
```
priitlatt, line 108: Maybe add an example?

```suggestion
            'separated by period in the format [Major].[Minor].[Patch]. '
            'For example `3.2.46`'
```

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +97,89 @@ class AppStoreConnectArgument(cli.Argument):
...
+        description='Build processing state',
+        argparse_kwargs={
+            'required': False,
+            'choices': list(BuildProcessingState),
+        },
+    )
+    BUILD_VERSION_NUMBER = cli.ArgumentProperties(
+        key='build_version_number',
+        flags=('--build-version-number',),
+        description='Build version number is the version number of the uploaded build',
```
priitlatt, line 178:
```suggestion
        description=(
            'Build version number is the version number of the uploaded build. '
            'For example `46`'
        ),
```

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -96,6 +97,89 @@ class AppStoreConnectArgument(cli.Argument):
...
+        key='processing_state',
+        flags=('--processing-state',),
+        type=BuildProcessingState,
+        description='Build processing state',
+        argparse_kwargs={
+            'required': False,
+            'choices': list(BuildProcessingState),
+        },
+    )
+    BUILD_VERSION_NUMBER = cli.ArgumentProperties(
```
priitlatt, line 175: Shouldn't the type be `int`?

priitlatt: Later in filter it will be converted to string anyway, but this preliminary validation would make it fail immediately on erroneous user input.

src/codemagic/cli/argument/argument.py
```diff
@@ -15,6 +15,17 @@
...
+    def resolve_optional_two_way_switch(cls,
+                                        is_switched_on: Optional[bool],
+                                        is_switched_off: Optional[bool]) -> Optional[bool]:
+        if {is_switched_on, is_switched_off} in ({True}, {False}):
+            raise ValueError('Neither of the switches, or exactly one can be truthy at the time')
+        if is_switched_on is True:
+            return True
+        if is_switched_off is True:
+            return False
+        return None
```
priitlatt, line 28:
```suggestion
        return None

```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+            self.echo(str(latest_build_number))
+            return latest_build_number
+
+    @cli.action('get-latest-app-store-build-number',
+                BuildArgument.APPLICATION_ID_RESOURCE_ID,
+                AppStoreVersionArgument.APP_STORE_VERSION)
+    def get_latest_app_store_build_number(self,
+                                           application_id: ResourceId,
+                                           app_store_version: Optional[str] = None,
+                                           should_print: bool = False) -> Optional[int]:
```
priitlatt, line 215:
```suggestion
                                          application_id: ResourceId,
                                          app_store_version: Optional[str] = None,
                                          should_print: bool = False) -> Optional[int]:
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -127,13 +133,20 @@ def _get_resource(self, resource_id, resource_manager, should_print):
         self.printer.print_resource(resource, should_print)
         return resource
 
-    def _list_resources(self, resource_filter, resource_manager, should_print):
+    def _list_resources(self, resource_filter, resource_manager, should_print, application_id=None):
```
priitlatt, line 136: This method does too much now for my taste. It would be better to separate listing for fixed resource and define new method which would be something like

```python
    def _list_resources_for(self, resource_id: ResourceId, resource_filter, resource_manager, should_print):
        try:
            resources, _included = resource_manager.list(resource_id, resource_filter=resource_filter)
        except AppStoreConnectApiError as api_error:
            raise AppStoreConnectError(str(api_error))
        ...
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -127,13 +133,20 @@ def _get_resource(self, resource_id, resource_manager, should_print):
...
+                resources = resource_manager.list(application_id, resource_filter)
+            else:
+                resources = resource_manager.list(resource_filter=resource_filter)
         except AppStoreConnectApiError as api_error:
             raise AppStoreConnectError(str(api_error))
 
-        self.printer.log_found(resource_manager.resource_type, resources, resource_filter)
+        self.printer.log_found(
+            resource_manager.resource_type,
+            resources if isinstance(resources, List) else resources[0],
```
priitlatt, line 147: It is not transparent at all why something that isn't list is accessed by numerical index. Also, `isinstance` check should be done with `list` type, and not with `List` type annotation type.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                    pre_release_version: Optional[str] = None,
+                    processing_state: Optional[BuildProcessingState] = None,
+                    build_version_number: Optional[str] = None,
+                    should_print: bool = True) -> List[Build]:
+        """
+        List Builds from Apple Developer Portal matching given constraints
+        """
+        try:
+            expired_value = Argument.resolve_optional_two_way_switch(expired, not_expired)
+        except ValueError as ve:
```
priitlatt, line 186: Unused variable
```suggestion
        except ValueError:
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+            flags = f'{BuildArgument.EXPIRED.flags[0]!r} and {BuildArgument.NOT_EXPIRED.flags[0]!r}'
```
priitlatt, line 187: Argument now has `flag` property, which returns the longest (as in most self explanatory and human-friendy) defined flag name
```suggestion
            flags = f'{BuildArgument.EXPIRED.flag!r} and {BuildArgument.NOT_EXPIRED.flag!r}'
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                BuildArgument.PROCESSING_STATE,
+                BuildArgument.BUILD_VERSION_NUMBER)
+    def list_builds(self,
+                    application_id: Optional[ResourceId] = None,
+                    expired: Optional[bool] = None,
+                    not_expired: Optional[bool] = None,
+                    build_id: Optional[ResourceId] = None,
+                    pre_release_version: Optional[str] = None,
+                    processing_state: Optional[BuildProcessingState] = None,
+                    build_version_number: Optional[str] = None,
```
priitlatt, line 179: I still find that this should come in as integer.
```suggestion
                    build_version_number: Optional[int] = None,
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+        return self._list_resources(builds_filter, self.api_client.builds, should_print)
+
+    def get_latest_build_number(self, builds: List[Build]) -> Optional[int]:
+        try:
+            latest_build_number = max((build.attributes.version for build in builds), key=int)
+        except ValueError:
+            return None
+        else:
+            self.echo(str(latest_build_number))
+            return latest_build_number
```
priitlatt, line 207: It actual return type doesn't match with type hint as `Build.Attributes.version` is declared as str.

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+            app=application_id,
+            expired=expired_value,
+            id=build_id,
+            processing_state=processing_state,
+            version=build_version_number,
+            pre_release_version_version=pre_release_version,
+        )
+        return self._list_resources(builds_filter, self.api_client.builds, should_print)
+
+    def get_latest_build_number(self, builds: List[Build]) -> Optional[int]:
```
priitlatt, line 200: Why is this method not private? And it could be declared as classmethod too.

src/codemagic/tools/app_store_connect.py
priitlatt, line 207: `else` block is rendundant since `except` clause terminates method anyway. 
```suggestion
        self.echo(str(latest_build_number))
        return latest_build_number
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +161,81 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+    def get_latest_app_store_build_number(self,
+                                           application_id: ResourceId,
+                                           app_store_version: Optional[str] = None,
+                                           should_print: bool = False) -> Optional[int]:
+        """
+        Get latest App Store build number for the given application
+        """
+        versions_client = self.api_client.app_store_versions
+        versions_filter = versions_client.Filter(version_string=app_store_version)
+        _, builds = self._list_resources(versions_filter, versions_client, should_print, application_id=application_id)
```
priitlatt, line 221: As stated above. A dedicated _list_ method should be declared for using with specific resource id.

priitlatt (changes requested):

src/codemagic/tools/app_store_connect.py
BondarenkoStas: it's converted to int in
```
max((build.attributes.version for build in builds), key=int)
```

src/codemagic/tools/app_store_connect.py
BondarenkoStas: as proposed, moved the logic out of `_list_resources` https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/87fa4db77f534bc9430952896e383ea509531914#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R215

src/codemagic/tools/app_store_connect.py
BondarenkoStas: not relevant anymore https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/87fa4db77f534bc9430952896e383ea509531914#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R137

src/codemagic/tools/app_store_connect.py
BondarenkoStas: done [here](https://github.com/codemagic-ci-cd/cli-tools/pull/67/commits/87fa4db77f534bc9430952896e383ea509531914#diff-c655fbe88a18ddd4d367405c06673842a69855a44563b37bd3b06625bf712419R194)

src/codemagic/apple/app_store_connect/resource_manager.py
```diff
@@ -74,6 +78,10 @@ def __init__(self, client: AppStoreConnectApiClient):
     def resource_type(self) -> Type[R]:
         raise NotImplemented
 
+    @classmethod
+    def _get_include_field_name(cls, include_type: Type[R]) -> str:
+        raise NotImplementedError()
```
priitlatt, line 83:
```suggestion
        raise NotImplemented
```
suits here better. `NotImplementedError` will trigger IDEs and linters to complain that classes which inherit from `Resource` should implement this abstract method, which is not the case.

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import TypeVar
+
+from codemagic.apple.app_store_connect.resource_manager import ResourceManager
+from codemagic.apple.resources import AppStoreVersion
+from codemagic.apple.resources import Build
+from codemagic.apple.resources import Resource
+from codemagic.apple.resources import ResourceId
+
+
+IncludedResource = TypeVar('IncludedResource', bound=Resource)
```
priitlatt, line 15:
```suggestion
IncludedResource = TypeVar('IncludedResource', bound=Resource)
```

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import TypeVar
+
+from codemagic.apple.app_store_connect.resource_manager import ResourceManager
+from codemagic.apple.resources import Build
+from codemagic.apple.resources import PreReleaseVersion
+from codemagic.apple.resources import Resource
+from codemagic.apple.resources import ResourceId
+
+
+IncludedResource = TypeVar('IncludedResource', bound=Resource)
```
priitlatt, line 15:
```suggestion
IncludedResource = TypeVar('IncludedResource', bound=Resource)
```

src/codemagic/apple/app_store_connect/versioning/pre_release_versions.py
```diff
@@ -0,0 +1,55 @@
...
+    @classmethod
+    def _get_include_field_name(cls, include_type: Type[IncludedResource]) -> str:
+        if include_type is Build:
+            return 'builds'
+        raise ValueError(f'Unknown include type {include_type}')
+
+    def list_with_include(
+        self,
+        include_type: Type[IncludedResource],
+        resource_filter: Filter = Filter()) -> Tuple[List[PreReleaseVersion], List[IncludedResource]]:
```
priitlatt, line 42: Formatting
```suggestion
    def list_with_include(
            self,
            include_type: Type[IncludedResource],
            resource_filter: Filter = Filter()) -> Tuple[List[PreReleaseVersion], List[IncludedResource]]:
```

src/codemagic/apple/app_store_connect/versioning/app_store_versions.py
```diff
@@ -0,0 +1,56 @@
...
+    def _get_include_field_name(cls, include_type: Type[IncludedResource]) -> str:
+        if include_type is Build:
+            return 'build'
+        raise ValueError(f'Unknown include type {include_type}')
+
+    def list_with_include(
+        self,
+        application_id: ResourceId,
+        include_type: Type[IncludedResource],
+        resource_filter: Filter = Filter()) -> Tuple[List[AppStoreVersion], List[IncludedResource]]:
```
priitlatt, line 42: Formatting:
```suggestion
    def list_with_include(
            self,
            application_id: ResourceId,
            include_type: Type[IncludedResource],
            resource_filter: Filter = Filter()) -> Tuple[List[AppStoreVersion], List[IncludedResource]]:
```

src/codemagic/tools/android_app_bundle.py
```diff
@@ -378,10 +378,7 @@ def is_signed(self, aab_path: pathlib.Path):
 
     @cli.action('validate', AndroidAppBundleArgument.BUNDLE_PATH)
     def validate(self, aab_path: pathlib.Path) -> str:
-        """
-        Verify that given Android App Bundle is valid and print
-        out information about it.
-        """
+        """ Verify that given Android App Bundle is valid and print out information about it """
```
priitlatt, line 381:
```suggestion
        """
        Verify that given Android App Bundle is valid and print out information about it
        """
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -12,12 +12,15 @@
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
+from typing import Union
```
priitlatt, line 15: Unused import
```suggestion
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -26,13 +29,17 @@
 from codemagic.apple.resources import Profile
 from codemagic.apple.resources import ProfileState
 from codemagic.apple.resources import ProfileType
+from codemagic.apple.resources import Resource
```
priitlatt, line 32: Unused import
```suggestion
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +155,91 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+            processing_state=processing_state,
+            version=build_version_number,
+            pre_release_version_version=pre_release_version,
+        )
+        return self._list_resources(builds_filter, self.api_client.builds, should_print)
+
+    @classmethod
+    def _get_latest_build_number(cls, builds: List[Build]) -> Optional[int]:
+        try:
+            latest_build_number = max((build.attributes.version for build in builds), key=int)
```
priitlatt, line 197: `Build.Attributes.version` is declared as string. To respect the return value it should be converted to integer here.
```suggestion
            latest_build_number = max(int(build.attributes.version) for build in builds)
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -148,6 +155,91 @@ def _delete_resource(self, resource_manager, resource_id: ResourceId, ignore_not
...
+                                          application_id: ResourceId,
+                                          app_store_version: Optional[str] = None,
+                                          should_print: bool = False) -> Optional[int]:
+        """
+        Get latest App Store build number for the given application
+        """
+        versions_client = self.api_client.app_store_versions
+        versions_filter = versions_client.Filter(version_string=app_store_version)
+        try:
+            _versions, builds = versions_client.list_with_include(application_id, Build, resource_filter=versions_filter)
```
priitlatt, line 216: Line too long
```suggestion
            _versions, builds = versions_client.list_with_include(
                application_id, Build, resource_filter=versions_filter)
```

priitlatt (changes requested):

priitlatt:
When build resources are printed out to terminal then icon asset tokens are not formatted:

```
-- Build --
Id: effa9358-6dd7-4e19-aa45-d90372d4c047
Type: builds
Expired: False
Icon asset token: ImageAsset(templateUrl='https://is3-ssl.mzstatic.com/image/thumb/Purple124/v4/b1/af/6f/b1af6fd1-a4c4-3458-25ba-caa91896bed4/AppIcon.png/{w}x{h}bb.{f}', height=1024, width=1024)
Min os version: 10.11
Processing state: VALID
Version: 2
Uploaded date: 2021-02-15 10:37:57-08:00
Expiration date: 2021-05-16 10:37:57-07:00
```

Any formatting is not applied to `ImageAsset` and as a result it doesn't look as good as it could. A simple improvement over current solution would be
```
-- Build --
Id: effa9358-6dd7-4e19-aa45-d90372d4c047
Type: builds
Expired: False
Icon asset token: 
        Template url: https://is3-ssl.mzstatic.com/image/thumb/Purple124/v4/b1/af/6f/b1af6fd1-a4c4-3458-25ba-caa91896bed4/AppIcon.png/{w}x{h}bb.{f}
        Height: 1024
        Width: 1024
Min os version: 10.11
Processing state: VALID
Version: 2
Uploaded date: 2021-02-15 10:37:57-08:00
Expiration date: 2021-05-16 10:37:57-07:00
```

This tiny modification to `Resource._format_attribute_value` will make it work as seen above (see the case where `value` is of type `DictSerializable`).
```python
class Resource(LinkedResourceData, metaclass=PrettyNameMeta):

    ...

    def _format_attribute_value(self, attribute_name: str, value: Any) -> Any:
        if self._hide_attribute_value(attribute_name):
            return '"..."'
        if isinstance(value, enum.Enum):
            return value.value
        if isinstance(value, DictSerializable):
            lines = '\n'.join(f'\t{self._format_attribute_name(k)}: {v}' for k, v in value.dict().items())
            return f'\n{lines}'
        return value
```

priitlatt (approved):
Maybe wait until next week with deployment.

PR #69, opened 2021-02-17
Improvement: Add option to specify allowed apps for keychain add-certificates

priitlatt:
**Add options to `keychain add-certificates` to specify which applications have access to the imported certificate without warning.**

Up until now certificates are added to specified keychain in a way that only `codesign` can access them without prompting UI warning. But to fulfill the needs for not only iOS developers it might be necessare to make certificates usable from headless session for other executables too, such as `productsign`, which is used to sign macOS applications. The apps that can access certificate in keychain can be configured during import via `-T` flag on `security import ...`. If one desires to make the certificate fully available for all applications without warning, then it can be imported to keychain using `security import` in conjunction with`-A` option. By default certificates are now added to keychain so that
- both `codesign` and `productsign` can access them without warnings,
- `-A` or `--allow-all-applications` can be used with `keychain add-certificates` to allow any application to access the imported key without warning,
- `-D` or `--disallow-all-applications` can be used with `keychain add-certificates` in order to prohibit all applications from accessing the imported key without warning.

**Modified action**:
- [`keychain add-certificate`](https://github.com/codemagic-ci-cd/cli-tools/blob/feature/keychain-add-certificate-allowed-apps/docs/keychain/add-certificates.md)

**Example executions of modified actions**

- Specify path to non-existent executable
```bash
> keychain add-certificates -c certificate.p12 --allow-app /path/to/non/existent/executable
usage: keychain add-certificates [-h] [--log-stream {stderr,stdout}] [--no-color] [--version] [-s] [-v] [-c certificate-path [certificate-path ...]] [--certificate-password CERTIFICATE_PASSWORD]
                                    [-a allowed-app [allowed-app ...]] [-A] [-D] [-p PATH]
keychain add-certificates: error: argument -a/--allow-app: Application "/path/to/non/existent/executable" does not exist or is not in PATH
```

- Allow all apps to use certificate without warnings
```bash
> keychain add-certificates -c certificate.p12 --allow-all-applications
Add certificates to keychain /private/var/folders/s_/1kqm5m1x0gg_r7yvbm2j1h3h0000gn/T/build_ylx26sua.keychain
Searching for files matching certificate.p12
Add certificate /Users/priit/Library/MobileDevice/Certificates/certificate.p12 to keychain /private/var/folders/s_/1kqm5m1x0gg_r7yvbm2j1h3h0000gn/T/build_ylx26sua.keychain
1 key imported.
1 certificate imported.
```

- Error handling on conflicting options
```bash
> keychain add-certificates -c certificate.p12 --allow-all-applications --disallow-all-applications
usage: keychain [-h] [--log-stream {stderr,stdout}] [--no-color] [--version] [-s] [-v] {add-certificates,create,delete,get-default,initialize,list-certificates,lock,make-default,set-timeout,show-info,unlock} ...
keychain: error: argument -A/--allow-all-applications: Using mutually exclusive options '--allow-app' and '--disallow-all-applications'
```

### DIFF
--- a/src/codemagic/tools/keychain.py
+++ b/src/codemagic/tools/keychain.py
@@ -270,19 +270,6 @@ class Keychain(cli.CliApp, PathFinderMixin):
         with NamedTemporaryFile(prefix='build_', suffix='.keychain') as tf:
             self._path = pathlib.Path(tf.name)
 
-    @classmethod
-    def _get_certificate_allowed_applications(
-            cls, given_allowed_applications: Sequence[pathlib.Path]) -> Iterable[str]:
-        for application in given_allowed_applications:
-            resolved_path = shutil.which(application)
-            if resolved_path is None:
-                # Only raise exception if user-specified path is not present
-                if application not in KeychainArgument.ALLOWED_APPLICATIONS.get_default():
-                    raise KeychainArgument.ALLOWED_APPLICATIONS.raise_argument_error(
-                        f'Application "{application}" does not exist or is not in PATH')
-            else:
-                yield resolved_path
-
     @cli.action('add-certificates',
                 KeychainArgument.CERTIFICATE_PATHS,
                 KeychainArgument.CERTIFICATE_PASSWORD,
@@ -318,6 +305,19 @@ class Keychain(cli.CliApp, PathFinderMixin):
             raise KeychainError('Did not find any certificates from specified locations')
         for certificate_path in certificate_paths:
             self._add_certificate(certificate_path, certificate_password, add_for_all_apps, add_for_apps)
+
+    @classmethod
+    def _get_certificate_allowed_applications(
+            cls, given_allowed_applications: Sequence[pathlib.Path]) -> Iterable[str]:
+        for application in given_allowed_applications:
+            resolved_path = shutil.which(application)
+            if resolved_path is None:
+                # Only raise exception if user-specified path is not present
+                if application not in KeychainArgument.ALLOWED_APPLICATIONS.get_default():
+                    raise KeychainArgument.ALLOWED_APPLICATIONS.raise_argument_error(
+                        f'Application "{application}" does not exist or is not in PATH')
+            else:
+                yield resolved_path
 
     def _add_certificate(self,
                          certificate_path: pathlib.Path,
--- a/src/codemagic/tools/keychain.py
+++ b/src/codemagic/tools/keychain.py
@@ -6,6 +6,7 @@ import os
 import pathlib
 import shutil
 from tempfile import NamedTemporaryFile
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Sequence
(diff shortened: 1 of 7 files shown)


priitlatt:
@VeArnold, you might be interested in this.

src/codemagic/tools/keychain.py
```diff
@@ -273,13 +320,18 @@ def _add_certificate(self,
...
             '-f', "pkcs12",
             '-k', self.path,
-            '-T', shutil.which('codesign'),
             '-P', certificate_password.value,
-        ], obfuscate_patterns=obfuscate_patterns)
+        ]
+        if allow_for_all_apps:
+            import_cmd.append('-A')
+        for allowed_application in allowed_applications:
+            import_cmd.extend(['-T', str(allowed_application)])
```
BondarenkoStas, line 332: maybe change `str(allowed_applications)` to `shutil.which(allowed_applications)` just in case

src/codemagic/tools/keychain.py
```diff
@@ -77,6 +77,32 @@ class KeychainArgument(cli.Argument):
         description='Encrypted p12 certificate password',
         argparse_kwargs={'required': False, 'default': ''},
     )
+    ALLOWED_APPLICATIONS = cli.ArgumentProperties(
+        flags=('-a', '--allow-app'),
+        key='allowed_applications',
+        description='Specify an application which may access the imported key without warning',
```
BondarenkoStas, line 83: i'd do something like `Specify an application which may access the imported key without warning with its absolute path or with its executable name if it's in the Path.

src/codemagic/tools/keychain.py
```diff
@@ -245,24 +271,45 @@ def _generate_path(self):
...
 
-        self.logger.info(f'Add certificates to keychain {self.path}')
+        add_for_all_apps = False
+        add_for_apps: Sequence[pathlib.Path] = []
+        if allow_all_applications and disallow_all_applications:
+            raise KeychainArgument.ALLOW_ALL_APPLICATIONS.raise_argument_error(
+                f'Using mutually exclusive options '
+                f'{KeychainArgument.ALLOWED_APPLICATIONS.flag!r} and '
+                f'{KeychainArgument.DISALLOW_ALL_APPLICATIONS.flag!r}')
+        elif allow_all_applications:
```
BondarenkoStas, line 296: since previous `if` raises an exception, just `if` can be sufficient here

src/codemagic/tools/keychain.py
priitlatt: Defaults are now `codesign` and `productsign` which are resolved to full paths later during action invocation. So it should eat really anything and further description becomes redundant.

src/codemagic/tools/keychain.py
priitlatt: I prefer to keep all application spefic validations in a single block since conceptually they fit nicely together. Recent changes also make it slightly easier to read and follow.

src/codemagic/tools/keychain.py
priitlatt: With changes introduced in 8b8e19bb77e7a847662049d132a8df97ffaf826f all given application paths are either fully resolved, or an exception is thrown in case they are not found from path or they don't exist.

BondarenkoStas (approved):
LGTM

PR #71, opened 2021-02-19
Improvement/resolve warnings

priitlatt:
Some warnings are raised during [tests](https://github.com/codemagic-ci-cd/cli-tools/runs/1919403495?check_suite_focus=true#step:8:57):
- `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working`
- `PytestCollectionWarning: cannot collect test class 'TestSuite' because it has a __init__ constructor (from: tests/models/junit/test_xml_generation.py)`

Changes introduced in this pull request:
- Fix `Sequence` type annotation imports - instead of importing it from `collections` module, import it from `typing`.
- Add `__test__ = False` attribute to source classes that have name which starts with `Test`.

### DIFF
--- a/src/codemagic/models/junit/definitions.py
+++ b/src/codemagic/models/junit/definitions.py
@@ -21,6 +21,8 @@ class TestSuites:
     name: str
     test_suites: List[TestSuite] = field(default_factory=lambda: [])
 
+    __test__ = False  # Tell Pytest not to collect this class as test
+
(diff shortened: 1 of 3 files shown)


BondarenkoStas (approved):
LGTM

PR #76, opened 2021-03-02
Bugfix: Fix platform filter for listing bundle identifiers from App Store Connect

priitlatt:
App Store Connect's [List Bundle IDs](https://developer.apple.com/documentation/appstoreconnectapi/list_bundle_ids) API endpoint does not respect the value of `filter[platform]` query parameter and always returns bundle identifiers for all platforms. This leads to unexpected consequences and as such we need to do client side filtering in order to avoid this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@ Version 0.4.8
 -------------
 **Improvements**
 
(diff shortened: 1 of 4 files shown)


mikhail-tokarev (approved):
LGTM

PR #77, opened 2021-03-03
Bugfix/list bundle ids filters

priitlatt:
Changes introduced in pull request #76 were reverted in f7d31c9 due to too strict filters. Namely, filtering solely on given platform type is not optimal since type `UNIVERSAL` matches everything and bundle identifiers with this platform type should always be included in the listing. This pull request restores changes introduced in #76 but in a slightly loosened fashion so that said universal type would always be accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
+Version 0.4.9
+-------------
+**Improvements**
+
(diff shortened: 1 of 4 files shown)


mikhail-tokarev (approved):
LGTM

PR #80, opened 2021-03-15
Allow to not fail the `use-profiles` action when profile(s) can't be applied to a project(s)

BondarenkoStas:
Use case is when some `node_modules` or other dependency has an xcode project which can't be signed.
In such case instead of specifying the exact project, user can use the `--warn-only` flag.

Not outputting the exception to warning since it's quite long.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Version 0.5.1
+-------------
+
+**Improvements**
+
+- Feature: add the `warn-only` flag to `xcode-project use-profiles` to not fail the action when profile(s) can't be applied to an Xcode project(s)
+
 Version 0.5.0
 -------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Version 0.5.1
 
 **Improvements**
 
-- Feature: add the `warn-only` flag to `xcode-project use-profiles` to not fail the action when profile(s) can't be applied to an Xcode project(s)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+Version 0.5.1
+-------------
+
+**Improvements**
+
+- Feature: add the `warn-only` flag to `xcode-project use-profiles` to not fail the action when profile(s) can't be applied to an Xcode project(s)
```
priitlatt, line 6:
```suggestion
- Feature: add the `warn-only` flag to `xcode-project use-profiles` not to fail the action when profiles can't be applied to any of the Xcode projects
```

src/codemagic/tools/_xcode_project/arguments.py
```diff
@@ -83,6 +83,16 @@ class XcodeProjectArgument(cli.Argument):
...
         },
     )
+    WARN_ONLY = cli.ArgumentProperties(
+        key='warn_only',
+        flags=('--warn-only',),
+        type=bool,
+        description=(
+            'Show warning when profiles can not be applied to an Xcode project(s) '
+            'instead of failing the action'
+        ),
```
priitlatt, line 93:
```suggestion
        description=(
            'Show warning when profiles cannot be applied to any of the Xcode projects '
            'instead of fully failing the action'
        ),
```

priitlatt (changes requested):

priitlatt (approved):

PR #82, opened 2021-03-15
Include certificate common name in SigningCertificate string representation

priitlatt:
WIP

### DIFF
--- a/src/codemagic/apple/resources/signing_certificate.py
+++ b/src/codemagic/apple/resources/signing_certificate.py
@@ -4,7 +4,10 @@ from base64 import b64decode
 from dataclasses import dataclass
 from dataclasses import field
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


priitlatt:
I could have used `Certificate` from `models` package to extract common name, but maybe in the future we'd like to split App Store Connect API client into standalone Python package and then it would introduce issues and as such I made this slight feature duplication.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@ Version 0.5.1
```
BondarenkoStas, line 3: I think 
```
Version 0.5.2
-------------
```
is missing here

priitlatt: You are absolutely correct.

BondarenkoStas (approved):

PR #86, opened 2021-03-18
Dependencies: Use PyJWT 2.0.1

priitlatt:
[PyJWT](https://pyjwt.readthedocs.io/en/stable/) version 2.0.0 introduced tiny change in the API that `jwt.encode` started returning `str` instead of `bytes` as in earlier versions. This caused a regression that was fixed in [v0.4.2](https://github.com/codemagic-ci-cd/cli-tools/releases/tag/v0.4.2) by freezing PyJWT version to 1.7.1.

To be compatible with the API change, handle both `str` and `bytes` type tokens in App Store Connect API client for backwards compatibility. But not to lag behind with outdated dependencies, bump PyJWT dependency to 2.0.0+ in requirements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Version 0.5.3
+-------------
+
+**Dependencies**
+
+- Update [PyJWT](https://pyjwt.readthedocs.io/en/stable/) Python dependency to version ~=2.0.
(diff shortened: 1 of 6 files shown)


BondarenkoStas (approved):
LGTM

PR #114, opened 2021-06-16
Update error messages

craastad:
These error messages are returned when `--certificate-key` is not provided. This is really easy to confuse with App Store Connect `--private-key`, which can mislead users.

I've also updated the README to make it clear these tools are useful outside of codemagic.io.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Version 0.8.2
+-------------
+
+**Improvements**
+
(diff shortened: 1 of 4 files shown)


priitlatt (changes requested):
@craastad, very nice changes. But please also update version in [`__version__.py`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/src/codemagic/__version__.py) and list out changes in [`CHANGELOG.md`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/CHANGELOG.md).

priitlatt (approved):
All good now.

Issue #126, opened 2021-06-28
Do not create keychains to temporary directory by default


craastad:
Naively using CLI tools can have undesirable consequences for the mac keychain. We should work to avoid these for default usages.

My mistake was fixed with `keychain make-default ~/Library/Keychains/login.keychain-db`

> I think that maybe we shouldn’t generate the keychain into $TMPDIR directory by default. when default keychain is not restored back to the system default login keychain and a reboot is performed, then the previous default keychain goes missing and system OS finds itself in a odd situation where default keychain is set, but it doesn’t exist.

The code changes are drafted in: https://github.com/codemagic-ci-cd/cli-tools/pull/125/files

priitlatt:
Currently [`keychain initialize`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/keychain/initialize.md) indeed creates new keychain into a temporary directory by default. This can cause issues if system is rebooted and login keychain is not set to default beforehand.

PR #129, opened 2021-07-02
Bugfix: keychain created by `keychain initialize` is inaccessible by Xcode

priitlatt:
The default keychain path chage from PR #126 broke iOS builds executed with `xcode-project build-ipa` as Xcode is unable to access secrets from the keychain created in `~/Library/Keychains/<keychain-name>.keychain-db`. Patch this by creating keychains in `~/Library/codemagic-cli-tools/keychains` by default using `keychain initialize`.


artemii-yanushevskyi:
Just to confirm, how could we be sure that `~/Library/codemagic-cli-tools/keychains` will be accessible by Xcode?

priitlatt:
https://github.com/codemagic-ci-cd/cli-tools/pull/129#issuecomment-872926507
@artemii-yanushevskyi, I tested it properly now 😉

artemii-yanushevskyi (approved):

PR #142, opened 2021-08-12
Fix: missing submittedDate attribute

artemii-yanushevskyi:


src/codemagic/apple/resources/beta_app_review_submission.py
```diff
@@ -15,6 +17,7 @@ class BetaAppReviewSubmission(Resource):
     @dataclass
     class Attributes(Resource.Attributes):
         betaReviewState: BetaReviewState
+        submittedDate: Optional[datetime]
```
priitlatt, line 20: If this field is defined in the API response, it will not become datetime instance here since no conversion is performed. See how dates are handled on `Profile` resource for example: https://github.com/codemagic-ci-cd/cli-tools/blob/v0.9.3/src/codemagic/apple/resources/profile.py#L35

priitlatt (changes requested):
Changelog and version update are missing.

priitlatt (approved):

PR #153, opened 2021-09-16
Feature: Xcode archive errors

priitlatt:
In case running `xcode-project build-ipa` fails with `xcodebuild` exception during archiving, then xcpretty formatter (which is enabled by default) usually hides underlying error, and as a result it is difficult to find out what was the actual reason for failure. For example:
```
$ xcode-project build-ipa --workspace my-project.xcworkspace --scheme "my scheme" --config Release
...

▸ Check Dependencies
Failed to archive my-project.xcworkspace
```
is not very informative.

The changes introduced in this PR try to solve this issue by detecting the failed archiving steps from unformatted `xcodebuild` logs, and then showing them in the very end of the build log. For example:
```shell
$ xcode-project build-ipa --workspace my-project.xcworkspace --scheme "my scheme" --config Release
...

▸ Check Dependencies
Failed to archive my-project.xcworkspace. The following build commands failed:

Check dependencies
	Code Signing Error: No signing certificate "iOS Distribution" found:  No "iOS Distribution" signing certificate matching team ID "X7NNQ9KYM5" with a private key was found.
	Code Signing Error: Provisioning profile "Development Profile" expired on 27. Aug 2021.
	Code Signing Error: Code signing is required for product type 'Application' in SDK 'iOS 14.5'
```


src/codemagic/utilities/backwards_file_reader.py
```diff
@@ -0,0 +1,39 @@
...
+            buffer = fd.read(min(unprocessed_size, buffer_size))
+            unprocessed_size -= buffer_size
+            lines = buffer.splitlines()
+
+            if buffer.endswith('\n'):  # Previous segment was not a half line.
+                yield current_segment or ''
+            else:  # Previous segment did not end at a line break.
+                lines[-1] += current_segment or ''
+
+            # Retain the first line for next iteration as it might might have some
```
artemii-yanushevskyi, line 32:
```suggestion
            # Retain the first line for next iteration as it might have some
```

artemii-yanushevskyi (approved):

PR #165, opened 2021-11-17
Feature: Add ANY_APP_STORE profile to app-store-connect CLI

dan-os:
Hello from the University of Southampton! 👋

In some cases, users may want to use an Apple "Distribution" certificate (instead of an "iOS Distribution" certificate) to sign their iOS apps. "Distribution" certificates can be used to sign any type of application (iOS, tvOS, Mac, Universal, etc.). A common reason users may want to do this is so that multiple app types can be built and signed using the same signing certificate, all from one workflow.

- This commit adds a new profile type to the "app-store-connect" CLI called "ANY_APP_STORE". This profile type simply tells the CLI to explicitly use a "Distribution" certificate. That's it!

Lastly, I'd like to thank all the maintainers for their hard work at Codemagic making our CI lives easy 🚀!

Best wishes,

Dan


priitlatt:
Hi @dan-os! Thank you for your contribution. Could you please bump the version in [`__version__.py`](https://github.com/dan-os/cli-tools/blob/feature/app-store-connect-distribution/src/codemagic/__version__.py#L3) by adding one to the patch portion of the version and document the changes you've made in [`CHANGELOG.md`](https://github.com/dan-os/cli-tools/blob/feature/app-store-connect-distribution/CHANGELOG.md)?

Also, could you please enlighten me how does one generate a profile with this type?

dan-os:
> Also, could you please enlighten me how does one generate a profile with this type?

This profile can be applied by specifying `--type` when invoking `app-store-connect`.

For example, to fetch the signing files for a bundle ID, the user might use:
```console
app-store-connect fetch-signing-files "$BUNDLE_ID" --type ANY_APP_STORE --create
```

dan-os:
I see! Apologies for not spotting this. I'll take another look to see if I can come up with an elegant solution. For now, I'll close this PR!

priitlatt:
@dan-os, the issue you tried to resolve with this PR still persist. Could you please open an issue for that so it will be on our radar?

Issue #176, opened 2021-12-20
App Store Connect API client JWT expiration


priitlatt:
~~As of version 0.14.1 the default expiration duration of App Store Connect API JSON Web Tokens used by used by [`AppStoreConnectApiClient`](https://github.com/codemagic-ci-cd/cli-tools/blob/v0.14.1/src/codemagic/apple/app_store_connect/api_client.py#L47) is just 30 seconds [[src](https://github.com/codemagic-ci-cd/cli-tools/blob/v0.14.1/src/codemagic/apple/app_store_connect/api_client.py#L82)], while it can be up to 20 minutes in case scope is not specified, and even longer if token scope is defined. See [App Store Connect API docs about generating tokens for API requests](https://developer.apple.com/documentation/appstoreconnectapi/generating_tokens_for_api_requests).~~

In some cases App Store Connect API respons with `401 Unauthorized` when completely valid JWT is provided in request headers, and this can make `app-store-connect` commands fail unexpectedly. We should reuse JWTs as long as possible to avoid such cases. In order to achieve this:
1. ~~increase the expiration time for generated JWTs from 30 seconds to something close to 20 minutes~~,
2. cache generated JWTs to disk so that tokens can be reused between CLI invocations.

priitlatt:
Actually JSON web token expiration is set to 19 minutes from the time they are created, and then they are discarded once less than 30 seconds is left until the hard expiration.

jleach:
I may have gone this yesterday also, after successfully uploading, the following API call failed. I though it was because my token had `Developer` access but according to the docs it should be enough. In any case, upload worked, but the `GET` call to `v1/apps` failed.

```
GET https://api.appstoreconnect.apple.com/v1/apps?limit=100&sort=name&filter%5BbundleId%5D=xxx.xxx.xxx.TestBuildAndSign returned 401: Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens
```

priitlatt:
Will be fixed by #184.

PR #178, opened 2021-12-21
Add option to retry unauthorized App Store Connect requests

priitlatt:
Sometimes App Store Connect API yields false positive results with regards to authentication. More precisely, requests signed with JSON web tokens that are generated from completely valid API keys can result in a response with status code `401` and content
```
Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens
```

To mitigate this issue to some degree, add option to gracefully retry requests that failed with status code 401.

**Affected actions:**
- All actions and subactions in `app-store-connect`.

**For example:**
```shell
$ app-store-connect fetch-signing-files io.example.app \
    --issuer-id 'Invalid API key issuer ID' \
    --api-unauthorized-retries 2 \
    --log-api-calls
>>> GET https://api.appstoreconnect.apple.com/v1/bundleIds {'limit': 100, 'sort': 'name', 'filter[identifier]': 'io.example.app', 'filter[platform]': 'IOS'}
<<< 401 {'errors': [{'status': '401', 'code': 'NOT_AUTHORIZED', 'title': 'Authentication credentials are missing or invalid.', 'detail': 'Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens'}]}
Request failed due to authentication failure on attempt #1, try again
>>> GET https://api.appstoreconnect.apple.com/v1/bundleIds {'limit': 100, 'sort': 'name', 'filter[identifier]': 'io.example.app', 'filter[platform]': 'IOS'}
<<< 401 {'errors': [{'status': '401', 'code': 'NOT_AUTHORIZED', 'title': 'Authentication credentials are missing or invalid.', 'detail': 'Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens'}]}
Unauthorized request retries are exhausted with 2 attempts, stop trying
[...]

PR #181, opened 2021-12-22
Feature: Cache generated App Store Connect JSON web tokens to disk

priitlatt:
Sometimes App Store Connect API yields false positive results with regards to authentication. More precisely, requests signed with JSON web tokens that are generated from completely valid API keys can result in a response with status code 401 and content
```
Authentication credentials are missing or invalid. - Provide a properly configured and signed bearer token, and make sure that it has not expired. Learn more about Generating Tokens for API Requests https://developer.apple.com/go/?id=api-generating-tokens
```
To reduce the chance of this to happen, one way it to add a graceful retry logic (as is done in #178). Another way is to use JSON Web Tokens that are known to work as many times as possible. The changes in this pull request implement the latter approach.

By default App Store Connect API requests initiated by `app-store-connect` actions try to load JSON Web Token from a file cache instead of generating new token for each individual command invocation. This can be turned off by specifying `--disable-jwt-cache` option.

**Affected actions:**
- All actions and subactions in app-store-connect.

**Examples**

- Caching JWT is enabled by default:
```shell
$ app-store-connect apps get 1481211155 --verbose
[12:21:00] INFO  > Get App 1481211155
[12:21:00] DEBUG > Load JWT for App Store Connect from disk cache
[12:21:00] DEBUG > Loaded JWT for App Store Connect from disk cache
-- App --
Id: 1481211155
...
```

- If cache does not exist, new token is generated
```shell
$ app-store-connect apps get 1481211155 --verbose
[12:24:29] INFO  > Get App 1481211155
[12:24:29] DEBUG > Load JWT for App Store Connect from disk cache
[12:24:29] DEBUG > Failed to load App Store Connect JWT from disk cache: Token is not cached
[12:24:29] DEBUG > Generate new App Store Connect JWT
[12:24:29] DEBUG > Cached App Store Connect JWT
-- App --
Id: 1481211155
...
```

- Disabled JWT caching
```shell
$ app-store-connect apps get 1481211155 --verbose --disable-jwt-cache
[12:21:07] INFO  > Get App 1481211155
[12:21:07] DEBUG > Failed to load App Store Connect JWT from disk cache: Disk cache is disabled
[12:21:07] DEBUG > Generate new App Store Connect JWT
-- App --
Id: 1481211155
...
```

Docs related to the changes made here are in a separate PR #183.


priitlatt:
Fixes #176 .

artemii-yanushevskyi (approved):

Issue #158, opened 2021-10-29
Signing Certificate type `IOS_DISTRIBUTION` doesn't fall back to `DISTRIBUTION`


artemii-yanushevskyi:
I created `DISTRIBUTION` Signing Certificate and a corresponding provisioning profile. When I try to fetch signing files I get the following error

```
app-store-connect fetch-signing-files --platform IOS --type IOS_APP_STORE --strict-match-identifier io.codemagic.capybara

Found 1 Bundle ID matching specified filters: identifier=io.codemagic.capybara, platform=IOS.
Found 3 Signing Certificates matching specified filters: certificateType=IOS_DISTRIBUTION.
Did not find any Signing Certificates for given private key
Did not find IOS_DISTRIBUTION Signing Certificates
```

I would expect that `DISTRIBUTION` Signing Certificates will be fetched if `IOS_DISTRIBUTION` Signing Certificate were not found

PR #185, opened 2021-12-28
Feature: Use "Distribution" certificate for iOS App Store and Mac App Store profiles

priitlatt:
**Background information**

Xcode 11 introduced a new kind of signing certificate _Apple Distribution_ with type `DISTRIBUTION`. It allowed using a single certificate to code sign all types of applications (iOS, tvOS, MacOS, etc), as opposed to the previous approach which required dedicated distribution certificate for each target platform. Until now provisioning profiles of type `IOS_APP_STORE` and `MAC_APP_STORE` were assigned a certificate with type `IOS_DISTRIBUTION` and `MAC_APP_DISTRIBUTION`, respectively, when downloaded (or created) using `app-store-connect fetch-signing-files`. This prevented reusing one signing certificate for signing apps that target different platforms, and created excess certificates in Apple Developer Portal.

Changes here update default provisioning profile type to certificate type mapping in the following way:
- `IOS_APP_STORE` provisioning profiles use `DISTRIBUTION` certificates by default (previously `IOS_DISTRIBUTION`),
- `MAC_APP_STORE` provisioning profiles use `DISTRIBUTION` certificates by default (previously `MAC_APP_DISTRIBUTION`).
Note that the change is **backwards compatible** in the sense that existing matching `IOS_DISTRIBUTION` and `MAC_APP_DISTRIBUTION` certificates are still used for respective provisioning profiles in case they are found and no `DISTRIBUTION` certificates are present.

Induced changes include possibility to list certificates using `app-store-connect list-certificates` by specifying more than certificate type by `--type` option.

There is also **one deprecation** in the Python API. First argument of `AppStoreConnect.list_certificates` was changed from
```python
certificate_type: Optional[CertificateType] = None
```
to
```python
certificate_types: Optional[Union[CertificateType, Sequence[CertificateType]]] = None
```
The **change is fully backwards compatible** for now. Positional usage works exactly as it used to, and a warnign is logged when the deprecated keyword argument is used.

**Updated actions:**
- `app-store-connect fetch-signing-files`,
- `app-store-connect list-certificates`.


priitlatt:
Follow up to #165.

priitlatt:
Keeping @dan-os in the loop, who originally pointed this out.

priitlatt:
Fixes #158.

artemii-yanushevskyi (approved):

src/codemagic/apple/resources/enums.py
```diff
@@ -190,13 +189,13 @@ class CertificateType(ResourceEnum):
...
         if profile_type is profile_type.IOS_APP_ADHOC:
             return CertificateType.IOS_DISTRIBUTION
         elif profile_type is profile_type.IOS_APP_DEVELOPMENT:
             return CertificateType.IOS_DEVELOPMENT
         elif profile_type is profile_type.IOS_APP_STORE:
-            return CertificateType.IOS_DISTRIBUTION
+            return CertificateType.DISTRIBUTION
         elif profile_type is profile_type.MAC_APP_DEVELOPMENT:
             return CertificateType.MAC_APP_DEVELOPMENT
         elif profile_type is profile_type.MAC_APP_STORE:
```
artemii-yanushevskyi, line 201: macOS is still using `MAC_APP_DISTRIBUTION` by default. It's better to use `DISTRIBUTION` by default instead

artemii-yanushevskyi (changes requested):

src/codemagic/apple/resources/enums.py
priitlatt: Resolved in c2686c0. PR description and changelog have also recieved relevant updates.

artemii-yanushevskyi (approved):

PR #187, opened 2021-12-28
Feature: Add interface to declare aliases for legacy actions

priitlatt:
At the moment there is no way to gracefully change tool namespace. For example the root namespace of `app-store-connect` is utterly cluttered and many of the actions could and should be grouped under relevant action groups.

For example at the moment we have action `app-store-connect get-profile`. But as there are a lot more profile actions, then it would make sense to create a separate action group `profiles`, and make the `get-profile` action under that as `app-store-connect profiles get`. Just doing a hard rename would be a breaking change that we want to avoid at all cost.

This PR adds option to add a deprecated alias via `deprecated_alias` parameter to `@cli.action` decorator. By moving forwards with the example above, with the changes from this PR we can refactor
```python
    @cli.action('get-profile', *GetProfileArgs)
    def get_profile(...): ...
```
from `AppStoreConnect` class to `ProfilesActionGroup` as
```python
@cli.action(
    'get',
    *GetProfileArgs,
    action_group=AppStoreConnectActionGroup.PROFILES,
    deprecated_alias='get-profile',
)
def get_profile(...): ...
```
Then both action `app-store-connect get-profile` and `app-store-connect profiles get` would do the same thing with the difference that the deprecated version does not pop up in `--help` messages and yields a warning message:

```shell
Using `app-store-connect get-profile` is deprecated and replaced by equivalent action `app-store-connect profiles get`.
Use `app-store-connect profiles get` instead as `app-store-connect get-profile` is subject for removal in future releases.
```


artemii-yanushevskyi (changes requested):
It seems like we use "legacy" and "deprecated" interchangeably, can we perhaps stick to one or the other?

src/codemagic/cli/cli_help_formatter.py
```diff
@@ -14,14 +18,28 @@ def __init__(self, *args, **kwargs):
         else:
             self._width = sys.maxsize
 
+    @classmethod
+    def suppress_action(cls, action_name: str):
+        cls._suppressed_actions.add(action_name)
+
+    def _exclude_suppressed_actions(self, message: str) -> str:
```
artemii-yanushevskyi, line 25: I was wondering if there is a way to avoid adding deprecated actions, so that we don't have to suppress them later, if not, let's add a comment justifying the usage of this function.

src/codemagic/cli/cli_help_formatter.py
```diff
@@ -1,11 +1,15 @@
...
+import re
 import shutil
 import sys
+from typing import Set
 
 from .colors import Colors
 
 
 class CliHelpFormatter(argparse.HelpFormatter):
+    _suppressed_actions: Set[str] = set()
```
artemii-yanushevskyi, line 11:
```suggestion
    _deprecated_actions: Set[str] = set()
```

priitlatt: Addressed in [a821077](https://github.com/codemagic-ci-cd/cli-tools/pull/187/commits/a821077e1e61d59a4e4c763bdedacb8e0a4c4ce9).

src/codemagic/cli/cli_help_formatter.py
priitlatt: We still need to add those actions so that `argparse` would be aware of them. Otherwise we'd fail prematurely with
```shell
<tool-name> error: argument action: invalid choice: 'deprecated-action-name' (choose from ...)
```

PR #182, opened 2021-12-23
Add context to beta test group arguments

jleach:
Added some context to the beta test group so its more clear when it can be used.

PR #189, opened 2022-01-04
Add context to app store connect docs

jleach:
Added some context to the beta test group so its more clear when it can be used.

Replaces PR #182

### DIFF
--- a/src/codemagic/tools/_app_store_connect/arguments.py
+++ b/src/codemagic/tools/_app_store_connect/arguments.py
@@ -649,7 +649,11 @@ class PublishArgument(cli.Argument):
         key='submit_to_testflight',
         flags=('--testflight', '-t'),
         type=bool,
-        description='Enable submission of an app for Testflight beta app review to allow external testing.',
+        description=(
+            'Enable submission of an app for Testflight beta app review to allow '
+            'external testing. NOTE: Requires an API key with `App Manager` or '
+            'greater access. '
+        ),
         argparse_kwargs={
             'required': False,
             'action': 'store_true',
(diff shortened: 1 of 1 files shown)


src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -939,7 +943,13 @@ class BuildArgument(cli.Argument):
         key='beta_group_names',
         flags=('--beta-group',),
         type=str,
-        description='Name of your Beta group',
+        description=(
+            'Name(s) of your **external** beta test group. Internal test '
+            'groups have access to all builds so this arguments cannot '
+            'be used with them. If you use this for internal teams it will '
+            'upload correctly, but the command will fail to make the group '
+            'assignment. Multiple arguments accepted.'
```
priitlatt, line 951: Markdown formatting, such as `**external**` does not look nice in terminal help messages, and as such I suggest to use the `Colors` class instead on the messages. Current [master](https://github.com/codemagic-ci-cd/cli-tools/tree/master) has changes from #190, which transforms ANSI encoded bold portions in docstring to Markdown bold. As an example check the bold deprecation warning on `--skip-package-validation` option [docs](https://github.com/codemagic-ci-cd/cli-tools/blob/69548fec08c6e822badd3144fec3b73e462850d3/docs/app-store-connect/publish.md#--skip-package-validation--sv) and [declaration](https://github.com/codemagic-ci-cd/cli-tools/blob/69548fec08c6e822badd3144fec3b73e462850d3/src/codemagic/tools/_app_store_connect/arguments.py#L750).

Another thing is that the `--beta-group` flag is not used only for `app-store-connect publish`, but it has other use cases in [`app-store-connect beta-groups`](https://github.com/codemagic-ci-cd/cli-tools/blob/v0.16.1/docs/app-store-connect/beta-groups.md) actions. As such the part
> If you use this for internal teams it will upload correctly, but the command will fail to make the group assignment.
can be out of place or even misleading.

All in all, I suggest the following change:
```suggestion
            f'Names of your {Colors.BOLD("external")} beta test groups. '
            'Not to be confused with internal test groups that have access '
            'to all builds. Multiple arguments accepted.'
```

priitlatt (changes requested):
Please also update **UNRELEASED** section in [changelog](https://github.com/codemagic-ci-cd/cli-tools/blob/master/CHANGELOG.md#unreleased) with what was done in this PR.

priitlatt:
No activity.

PR #203, opened 2022-02-21
Improvement: `xcode-project use-profiles` logs for set code signing settings

priitlatt:
Sometimes building Xcode projects with `xcode-project build-ipa` fail with errors such as
```bash
error: TargetName has conflicting provisioning settings. TargetName is automatically signed for development, but a conflicting code signing identity iPhone Distribution has been manually specified. Set the code signing identity value to "Apple Development" in the build settings editor, or switch to manual signing in the Signing & Capabilities editor. (in target 'TargetName' from project 'Runner')
```
even though before running the build, action`xcode-project use-profiles` was called, and the code signing settings were seemingly successfully configured. 

Usually the reason for this is that `xcode-project use-profiles` did not assign usable provisioning profile to the Xcode target `TargetName`, which is actually required for successful Xcode project build.

The aim of this PR is to make it clearer when such conflicting code signing settings happen. This is achieved by two steps.

First make it possible to further debug what settings were configured for each found Xcode project target by logging `code_signign_manager.rb` output always to debug log file. Also make it possible to include this output to `xcode-project use-profiles` output via new command line switch `--code-signing-setup-verbose-logging` without the need to enable fully verbose log output.

Second and perhaps more notable change for end users is that `xcode-project use-profiles` now shows warnings about targets for which code signing settings were not updated. The warning are based on targets for which code signing settings were set to avoid excess noise.

<details><summary>Updated logs</summary> 
<p>
Screenshot 2022-02-21 at 15 33 59
</p>
</details>

**Updated actions:**
- `xcode-project use-profiles`.


mikhail-tokarev (approved):
LGTM

PR #226, opened 2022-05-15
Development: Add default values to all  arguments in `add_certificates` and `use_profiles` declarations

DetachHead:
this makes the api behave more like the cli and makes it much easier to use.

currently in my code i have to do this, which isn't ideal and adds unnecessary bloat to my script:
```py
XcodeProject().use_profiles(
    XcodeProjectArgument.XCODE_PROJECT_PATTERN.get_default(),
    XcodeProjectArgument.PROFILE_PATHS.get_default(),
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+Unreleased
+-------------
+**Development**
+- Use default values for all arguments in `Keychain.add_certificates` and `XcodeProject.use_profiles`
+
 Version 0.26.0
 -------------
 
--- a/src/codemagic/tools/keychain.py
+++ b/src/codemagic/tools/keychain.py
@@ -307,7 +307,7 @@ class Keychain(cli.CliApp, PathFinderMixin):
                 KeychainArgument.DISALLOW_ALL_APPLICATIONS)
     def add_certificates(
             self,
-            certificate_path_patterns: Sequence[pathlib.Path],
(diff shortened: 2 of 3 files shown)


priitlatt (changes requested):
As these changes are not breaking and they make the usage of Python API a bit more convenient, then in principle I don't have anything to object. But please could you update [changelog](https://github.com/DetachHead/cli-tools/blob/Keychain.add_certificates_default/CHANGELOG.md) too with the changes you've made. Version header could be _Unreleased_ and the notes should go under development section.

priitlatt:
Please add PR description that outlies changes and the rationale for what was done.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Apple has deprecated the [Create an App Store Version Submission](https://develo
   - `confirm_review_submission`,
   - `create_review_submission`,
   - `create_review_submission_item`.
+- Use default values for all arguments in `Keychain.add_certificates` and `XcodeProject.use_profiles`
```
priitlatt, line 27: Version 0.26.0 is already released and we cannot include those changes there any more. Instead Create a new section at the top 
```
Unreleased
-------------

**Development**
- <Your changes here>
```

priitlatt (changes requested):

priitlatt (approved):
LGTM. Thank you for your contribution!

PR #245, opened 2022-07-01
Bugfix: Resolve bundle identifiers from Info.plist files for `xcode-project use-profiles`

priitlatt:
Mechanism of how bundle identifiers are detected from Xcode projects for `xcode-project use-profiles` was changed in version [v0.27.0](https://github.com/codemagic-ci-cd/cli-tools/releases/tag/v0.27.0). Namely the script [`code_signing_manager.rb`](https://github.com/codemagic-ci-cd/cli-tools/pull/232/files#diff-d49c5b1f0a323fabce10f56a5b783485494ae12650faa1db910a8870bfdeeeb4) was edited and sadly those edits introduced some regressions.

In case bundle identifier is defined in information property list (Info.plist) file, then it will not be resolved from target's build configuration settings. As a result we don't know what bundle identifier corresponds to given build configuration, which in turn means that we cannot assign proper code signing settings to this configuration. Eventually this leads to a failed iOS build as code signign settings are not properly configured.

With the changes here, in case `PRODUCT_BUNDLE_IDENTIFIER` cannot be resolved from build settings, then check `CFBundleIdentifier` value from build configuration's Info.plist file. The definitions there might contain variables that need to be resolved on their own. To do that, `VariableResolver` class was restored. There
variable substitution logic is almost a verbatim copy-paste from the [previously working revision](https://github.com/codemagic-ci-cd/cli-tools/pull/232/files#diff-d49c5b1f0a323fabce10f56a5b783485494ae12650faa1db910a8870bfdeeeb4L51) as it was extensively used and tested in the past. The old spahgetty variable value resolving mechanism was however replaced by [`Xcodeproj`](https://github.com/CocoaPods/Xcodeproj) built-in method [`XCBuildConfiguration#resolve_build_setting`](https://www.rubydoc.info/gems/xcodeproj/Xcodeproj%2FProject%2FObject%2FXCBuildConfiguration:resolve_build_setting).

**Updated actions:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+Version 0.29.1
+-------------
+
+This is a bugfix release including changes from [PR #245](https://github.com/codemagic-ci-cd/cli-tools/pull/245).
+
+**Bugfixes**
+- Fix `xcode-project use-profiles` for cases when bundle identifier is defined in information property list files. Add fallback bundle identifier detection from Info.plist file to code signing setup script if `PRODUCT_BUNDLE_IDENTIFIER` is not resolved from build configuration.
+
 Version 0.29.0
 -------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "codemagic-cli-tools"
-version = "0.29.0"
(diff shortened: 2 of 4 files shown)


remarkov (approved):
LGTM

PR #256, opened 2022-08-26
Improvement: Show error message if devices are not provided on development or ad-hoc profile creation

VeArnold:
When creating development and ad hoc provisioning profiles, Apple requires the existence of suitable test devices on the Apple Developer Portal.

Previous implementations did not check whether appropriate devices were included in the API call.

This lead to `409` responses from [App Store Connect API Create Profile](https://developer.apple.com/documentation/appstoreconnectapi/create_a_profile) endpoint. For example a `POST` request to the `https://api.appstoreconnect.apple.com/v1/profiles` endpoint with the following content:
```json
{
   "data":{
      "type":"profiles",
      "attributes":{
         "name":"profile name",
         "profileType":"IOS_APP_DEVELOPMENT"
      },
      "relationships":{
         "certificates":{
            "data":[
               {
                  "id":"13CO2HS7YE",
                  "type":"certificates"
               }
            ]
         },
         "devices":{},
         "bundleId":{
            "data":{
               "type":"bundleIds",
               "id":"22JL9M6NPV"
            }
         }
      }
   }
}
```

Results in the following response:
```json
{
  "errors" : [ {
    "id" : "97c1dc61-f853-43c1-9996-36997eff095d",
    "status" : "409",
    "code" : "ENTITY_ERROR.RELATIONSHIP.REQUIRED",
    "title" : "Relationship is required but was not provided.",
    "detail" : "The relationship 'devices' is required but was not provided with this request.",
    "source" : {
      "pointer" : "data/relationships/devices"
    }
  } ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 Version 0.31.3
 -------------
 
-**Development**:
-- Show error message when no matching test device is found in the Apple Developer Portal when creating an Ad Hoc or development provisioning profile.
+**Bugfixes**
+- Show error message when no matching test device is found in the Apple Developer Portal when creating an Ad Hoc or development provisioning profile. [PR #256](https://github.com/codemagic-ci-cd/cli-tools/pull/256)
 
 Version 0.31.2
 -------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 0.31.3
+-------------
+
+**Development**:
+- Show error message when no matching test device is found in the Apple Developer Portal when creating an Ad Hoc or development provisioning profile.
+
 Version 0.31.2
 -------------
 
--- a/src/codemagic/apple/app_store_connect/provisioning/profiles.py
+++ b/src/codemagic/apple/app_store_connect/provisioning/profiles.py
@@ -57,6 +57,21 @@ class Profiles(ResourceManager[Profile]):
         """
         if profile_type.devices_not_allowed() and devices:
             raise ValueError(f'Cannot assign devices to profile with type {profile_type}')
+        elif profile_type.devices_required() and not devices:
+            if profile_type.is_tvos_profile:
+                device_type = 'tvOS'
+            elif profile_type.is_macos_profile:
+                device_type = 'macOS'
+            elif profile_type.is_ios_profile:
+                device_type = 'iOS'
+            else:
+                raise ValueError(f'Device type for profile type {profile_type} is unknown')
+
+            raise ValueError(
+                f'Cannot create profile: Apple requires that you register at least one {device_type} '
+                f'testing device on the Apple Developer Portal to create a {profile_type} profile'
+            )
+
         if devices is None:
             devices = []
         attributes = {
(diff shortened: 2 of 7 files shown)


src/codemagic/apple/app_store_connect/provisioning/profiles.py
```diff
@@ -57,6 +57,21 @@ def create(self,
...
+                device_type = 'tvOS'
+            elif profile_type.is_macos_profile:
+                device_type = 'macOS'
+            elif profile_type.is_ios_profile:
+                device_type = 'iOS'
+            else:
+                raise ValueError(f'Device type for profile type {profile_type} is unknown')
+
+            raise ValueError(
+                f'Cannot create profile: Apple requires that you register at least one {device_type} '
```
priitlatt, line 71: Error message is misleading. We have no information here whatsoever to back up the claim that there are no devices registered in Apple Developer portal. What we know is that this profile creation request does not include any testing devices, while they are required for successful creation.

src/codemagic/apple/resources/enums.py
```diff
@@ -214,9 +218,9 @@ def is_tvos_profile(self) -> bool:
         return self.value.startswith('TVOS_')
 
     def devices_not_allowed(self) -> bool:
-        return not self.devices_allowed()
+        return not self.devices_required()
 
-    def devices_allowed(self) -> bool:
+    def devices_required(self) -> bool:
```
priitlatt, line 223: This rename is a breaking change, but we need to persist backwords compatibility. As such I suggest you add an alias with deprectaion warning:
```python
    def devices_allowed(self) -> bool:
        # Deprecation warning saying that this method was renamed to
        # "devices_required" in version x.y.z and is subject to be removed
        # in future releases.
        log.get_logger(self.__class__).warning(...)
        return self.devices_required()
```

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Version 0.31.3
+-------------
+
+**Development**:
```
priitlatt, line 4: Isn't is more of a bugfix?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+- Show error message when no matching test device is found in the Apple Developer Portal when creating an Ad Hoc or development provisioning profile.
```
priitlatt, line 5: Link to the PR from which the changes originate from.

priitlatt (changes requested):

src/codemagic/apple/app_store_connect/provisioning/profiles.py
VeArnold: Valid point, only concentrated on the case when we automatically try to create profiles (and therefore also fetch the device list ourselves). Will try to come up with something more covering yet still hint at a possible action.

src/codemagic/apple/resources/enums.py
```diff
@@ -214,9 +220,18 @@ def is_tvos_profile(self) -> bool:
...
-        return not self.devices_allowed()
+        return not self.devices_required()
 
     def devices_allowed(self) -> bool:
+        warning = (
+            'Deprecation warning! Method '
+            '"devices_allowed" is deprecated in favor of "devices_required" in version 0.31.3 '
+            'and is subject for removal in future releases.'
+        )
+        log.get_logger(self.__class__).warning(Colors.RED(warning))
```
priitlatt, line 231: Warning logs should be yellow if you really want to color them, but I don't think this will be necessary. But definitely don't use red, it is for exceptions.

priitlatt (changes requested):

priitlatt (approved):

PR #261, opened 2022-09-09
Bugfix: iOS unit tests signing configuration setup

priitlatt:
Fix configuring code signing settings with `xcode-project use-profiles` for iOS unit testing build targets.

It used to be that empty code signign settings were always assigned to Xcode build targets with product type `com.apple.product-type.bundle.unit-test`. This works as long as tests are launched on a simulator, but when a physical devices are targeted as test destination, then code signing becomes mandatory.

Unit testing targets do not take profiles, but require information about certificate which can be used for signing. Signing information is obtained eitther from
- a provisioning provile that matches exactly the bundle identifier of test target build configuration,
- if the former does not exist, reuse test target host application matching build configuration.

**Updated actions**


remarkov (approved):
lgtm

PR #266, opened 2022-09-28
Feature: log project build settings

artemii-yanushevskyi:
Run `xcodebuild -showBuildSettings` as part of `xcode-project build-ipa`. This will print out all relevant information about the project that is being built.

By default, the output is available in file CLI tools logs.

To show build settings in terminal window (or in Codemagic), use `xcode-project build-ipa` with `--verbose` flag.

*Motivation: help with debugging iOS build failures*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 0.32.3
+-------------
+
+**Features**:
+- `xcode-project build-ipa` command supports `--verbose` flag to output Xcode project build settings.
+
 Version 0.32.2
 -------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ Version 0.33.0
 -------------
 
 **Features**:
-- Run `xcodebuild -showBuildSettings` before `xcodebuild archive` as part of action `xcode-project build-ipa` to capture project settings. The output of build settings is hidden by default, but shown if `--verbose` flag is set. 
+- Add `xcode-project show-build-settings` action. It outputs Xcode project build setting using `xcodebuild -showBuildSettings` command.
+- Update `xcode-project build-ipa` action to run `xcodebuild -showBuildSettings` before `xcodebuild archive` to capture Xcode project build settings. The output of build settings is hidden by default, but shown if `--verbose` flag is set.
 
 Version 0.32.2
 -------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-Version 0.32.3
+Version 0.33.0
 -------------
 
 **Features**:
--- a/src/codemagic/models/xcodebuild.py
+++ b/src/codemagic/models/xcodebuild.py
@@ -36,7 +36,8 @@ class Xcodebuild(RunningCliAppMixin):
                  target_name: Optional[str] = None,
                  configuration_name: Optional[str] = None,
                  scheme_name: Optional[str] = None,
-                 xcpretty: Optional[Xcpretty] = None):
+                 xcpretty: Optional[Xcpretty] = None,
+                 verbose: bool = False):
         self.logger = log.get_logger(self.__class__)
         self.xcpretty = xcpretty
         self.workspace = xcode_workspace.expanduser() if xcode_workspace else None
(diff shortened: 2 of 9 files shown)


src/codemagic/models/xcodebuild.py
```diff
@@ -256,22 +256,32 @@ def test(self,
...
+                process = XcodebuildCliProcess(command, xcpretty=xcpretty, print_streams=print_streams)
                 cli_app.logger.info('Execute "%s"\n', process.safe_form)
                 process.execute().raise_for_returncode(include_logs=False)
             else:
                 subprocess.check_output(command)
-        except subprocess.CalledProcessError as cpe:
-            if ignore_error_code and ignore_error_code == cpe.returncode:
-                return
+        except subprocess.CalledProcessError:
             raise IOError(error_message, process)
```
priitlatt, line 285:
```suggestion
        except subprocess.CalledProcessError as cpe:
            raise IOError(error_message, process) from cpe
```

src/codemagic/models/xcodebuild.py
```diff
@@ -256,22 +256,32 @@ def test(self,
         error_message = f'Failed to test {self.workspace or self.project}'
         self._run_command(cmd, error_message)
 
+    def show_build_settings(self):
+        cmd = [*self._construct_base_command(None), '-showBuildSettings']
+        error_message = f'Failed to obtain build settings {self.workspace or self.project}'
```
priitlatt, line 261:
```suggestion
        error_message = f'Failed to show build settings {self.workspace or self.project}'
```

src/codemagic/models/xcodebuild.py
```diff
@@ -256,22 +256,32 @@ def test(self,
...
+
+        # avoid formatting build settings output
+        self._run_command(cmd, error_message, print_streams=print_streams, format_streams=False)
+
     def _run_command(self,
                      command: List[str],
                      error_message: str,
-                     ignore_error_code: Optional[int] = None):
+                     print_streams: bool = True,
+                     format_streams: bool = True):
```
priitlatt, line 273: I find `format_streams` as argument name to be slightly misleading. In case `XcodeBuild` instance does not have `xcpretty`, then there is no formatting even if `format_streams` is set to `True`. I would rename it to something like `allow_xcpretty_formatting` instead to reflect actual meaning.

src/codemagic/tools/xcode_project.py
```diff
@@ -214,6 +214,8 @@ def build_ipa(self,
         xcodebuild = self._get_xcodebuild(**locals())
         clean and self._clean(xcodebuild)
 
+        xcodebuild.show_build_settings()
```
priitlatt, line 217:
```suggestion
        self.logger.info(Colors.BLUE(f'Check {(xcodebuild.workspace or xcodebuild.xcode_project).name} build settings'))
        xcodebuild.show_build_settings()
```

Perhaps it makes sense to add human-readable log statement about what is going on. Without anything we have just 
```shell
Execute "xcodebuild -project /path/to/project.xcodeproj -scheme <scheme_name> -showBuildSettings"
```
in the very beginning, which does not look especially nice.

priitlatt (changes requested):
Changelog is not updated.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Version 0.32.3
+-------------
+
+**Features**:
+- `xcode-project build-ipa` command supports `--verbose` flag to output Xcode project build settings.
```
priitlatt, line 5:
```suggestion
- Run `xcodebuild -showBuildSettings` before `xcodebuild archive` as part of action `xcode-project build-ipa` to capture project settings. The output of build settings is hidden by default, but shown if `--verbose` flag is set. 
```

priitlatt (changes requested):
I suppose this change is significant enough to warrant a change of minor version, not only the patch.

Additionally:
- There could be cases when showing build settings is not desirable and as such I think we should have a way to disable it. 
- To follow already existing patterns (like with `clean`), then it would be nice to add a standalone action also for showing build settings, like `xcode-project show-build-settings`, that does only this one thing.

priitlatt (approved):

PR #269, opened 2022-09-30
Dependencies: Remove `pyOpenSSL` dependency

priitlatt:


src/codemagic/models/private_key.py
```diff
@@ -84,8 +70,25 @@ def from_openssh_key(cls, buffer: AnyStr, password: Optional[AnyStr] = None) ->
...
+        try:
+            cryptography_private_key = serialization.load_pem_private_key(cls._bytes(pem_key), _password)
+        except ValueError as ve:
+            if 'bad decrypt' in str(ve).lower():
+                log.get_file_logger(cls).exception('Failed to initialize private key: Invalid password')
+                raise ValueError('Invalid private key passphrase') from ve
+            else:
+                log.get_file_logger(cls).exception('Failed to initialize private key: Invalid PEM contents')
+                raise ValueError('Invalid private key PEM content') from ve
+        except TypeError as te:
```
artemii-yanushevskyi, line 84: Isn't this case covered by `ValueError`? Just wanted to clarify if both errors are possible for invalid passphrase.

artemii-yanushevskyi (changes requested):

src/codemagic/models/private_key.py
priitlatt: There is a difference in how `cryptograhy` and `OpenSSL` handle keys that are non-encrypted or signed with empty passwords. `crypto.load_privatekey` from `OpenSSL` treats those two the same, by raising `crypto.Error` in both cases. However, `serialization.load_pem_private_key` from `cryptograhpy` will raise `ValueError` in one case and `TypeError` in other, distinguishing those two cases.

artemii-yanushevskyi (approved):

PR #271, opened 2022-10-06
Docs: Show exactly how to pass the custom export options JSON string to use-profiles 

nevercode-kevin:


priitlatt (changes requested):
Markdown documentation files are auto-generated and should not be changed manually. The change should be made in the source files instead where the arguments are defined. In case of this particular option it would be [here](https://github.com/codemagic-ci-cd/cli-tools/blob/updates/use-profiles/src/codemagic/tools/_xcode_project/arguments.py#L161).

Once the source files are updated, then docs can be updated using script [`doc.py`](https://github.com/codemagic-ci-cd/cli-tools/blob/updates/use-profiles/doc.py).

priitlatt (approved):
This will do. I'll take care of the version updates and changelog entries myself.

PR #295, opened 2022-12-20
Bugfix: Python 3.11.1 compatibility

### LINKED ISSUE #293: Error running `keychan.initialize`
## Description
Suddenly the command `keychain initialize` started failing in our CI. I checked the libraries versions and tried to install previous one but also noticed that the `android_app_bundle.py` didn't change since 2021

Might be the version of Python? Any suggestions?

```
keychain initialize

  File "/Library/Frameworks/Python.framework/Versions/3.11/bin/keychain", line 5, in <module>
    from codemagic.tools import Keychain
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/codemagic/tools/__init__.py", line 1, in <module>
    from .android_app_bundle import AndroidAppBundle
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/codemagic/tools/android_app_bundle.py", line 35, in <module>
    class AndroidAppBundleArgument(cli.Argument):
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/codemagic/tools/android_app_bundle.py", line 65, in AndroidAppBundleArgument
    KEYSTORE_PATH_REQUIRED = KEYSTORE_PATH.duplicate(argparse_kwargs={'required': True})
                             ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'duplicate'
```

priitlatt:
Fixes #293.

The way how Python treats `NamedTuple`s in version 3.11.1 is slightly different to how it used to be before (including in 3.11.0).

Consider this snippet for example:

```python
# named_tuple_valued_enums.py

import enum
import typing


class NT(typing.NamedTuple):
    s: str
    i: int


class E(NT, enum.Enum):
    A = NT(s='a', i=1)
    B = NT(s='b', i=A.i)


if __name__ == '__main__':
    assert type(E.A.value) == NT
```

In Python versions <= 3.11.0 one could access the fields of already defined enumerations. In the example above `B` is essentially defined as `NT(s='b', i=1)`, but the value for `i` is taken from `A.i`, which is defined right above. However in Python 3.11.1 the type of `A` at that point is not `NT` as it used to be, but is plain `tuple` now. 

Consequently the snippet above runs fine with Python <= 3.11.0, but fail with the following `AttributeError` on Python 3.11.1:

```python
  File "/private/tmp/named_tuple_valued_enums.py", line 10, in <module>
    class E(NT, enum.Enum):
  File "/private/tmp/named_tuple_valued_enums.py", line 12, in E
    B = NT(s='b', i=A.i)
                    ^^^
AttributeError: 'tuple' object has no attribute 'i'
```

As CLI arguments are implemented as enumerations, and some of them are defined using other previously defined arguments, then this makes the package unusable on Python 3.11.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 0.36.5
+-------------
+
+**Bugfix**
+- Support Python 3.11.1 [PR #295](https://github.com/codemagic-ci-cd/cli-tools/pull/295).
+
 Version 0.36.4
 -------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "codemagic-cli-tools"
-version = "0.36.4"
+version = "0.36.5"
 description = "CLI tools used in Codemagic builds"
 readme = "README.md"
 authors = [
--- a/src/codemagic/__version__.py
+++ b/src/codemagic/__version__.py
@@ -1,5 +1,5 @@
 __title__ = 'codemagic-cli-tools'
 __description__ = 'CLI tools used in Codemagic builds'
-__version__ = '0.36.4.dev'
+__version__ = '0.36.5.dev'
 __url__ = 'https://github.com/codemagic-ci-cd/cli-tools'
 __licence__ = 'GNU General Public License v3.0'
--- a/src/codemagic/cli/argument/argument.py
+++ b/src/codemagic/cli/argument/argument.py
@@ -38,7 +38,10 @@ class Argument(ArgumentProperties, enum.Enum):
         for argument in arguments:
             if argument in exclude:
                 continue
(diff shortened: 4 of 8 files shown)


priitlatt:
Without the changes [CI job failed](https://github.com/codemagic-ci-cd/cli-tools/actions/runs/3740934842/jobs/6349896525)  on Python 3.11, but [was fine](https://github.com/codemagic-ci-cd/cli-tools/actions/runs/3740934842/jobs/6349896342) on Python 3.7. As can be seen from above, all checks pass with the modifications proposed in this PR.

VeArnold (approved):
LGTM 🚀

PR #302, opened 2023-01-20
Bugfix: Code signing issues with Xcode 14+

priitlatt:
When using Xcode version 14 or above specify xcarg `CODE_SIGN_STYLE=Manual` for `xcodebuild archive` command (executed as part of `xcode-project build-ipa` action) if none of the provisioning profiles that are used for code signing are managed by Apple. This forces Xcode (14+) to use provisioning profiles that are defined in export options property list when performing `xcodebuild -exportArchive`.

More info at can be seen from this [thread](https://forums.swift.org/t/xcode-14-beta-code-signing-issues-when-spm-targets-include-resources/59685/22).

**Updated actions**:
- `xcode-project build-ipa`


src/codemagic/models/xcodebuild.py
```diff
@@ -190,12 +206,18 @@ def clean(self):
...
+    def archive(
+        self,
+        export_options: ExportOptions,
+        archive_directory: pathlib.Path,
+        *,
+        xcargs: Optional[str] = None,
+        custom_flags: Optional[str] = None,
+        xcode: Optional[Xcode] = None,
+    ) -> pathlib.Path:
+        if xcode is None:
```
VeArnold, line 218: Could you elaborate on this?

We only call archive in one place, where we pass `xcode` as a parameter. And there, we get the `xcode` version by the same `Xcode.get_selected()` call, which, if all is well, should return the version - though in case of error, it could also return `None` - but I don't think we even reach this then. And even if we did, it seems weird that we rely on the same method, in that case, moments later.

I would say either this argument is not optional, or we could, in fact, not even attempt to pass it (and resolve it within the `archive` call). Currently, it just looks like double trouble. Am I missing something?

src/codemagic/models/xcodebuild.py
```diff
@@ -53,9 +58,11 @@ def _get_logs_path(self) -> pathlib.Path:
             tmp_dir = pathlib.Path(tempfile.gettempdir())
         logs_dir = tmp_dir / 'xcodebuild_logs'
         logs_dir.mkdir(exist_ok=True)
-        prefix = f'{self.xcode_project.stem}_'
-        with tempfile.NamedTemporaryFile(prefix=prefix, suffix='.log', dir=logs_dir) as tf:
-            return pathlib.Path(tf.name)
+        path_candidates = itertools.chain(
```
VeArnold, line 61: Is this change related?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+Version 0.37.1
+-------------
+
+TBD
```
VeArnold, line 4: Guess this was moved from `draft` to soon.

VeArnold (changes requested):

src/codemagic/models/xcodebuild.py
priitlatt: Not really. But it helped a great deal with testing everything out locally. You know the previous way of logfile generation resulted in random suffixes in file name that could not be easily sorted without taking file creation time into account.

VeArnold:
:shipit:

VeArnold (approved):

Issue #303, opened 2023-01-23
Action get-latest-testflight-build-number returns version number instead of build number


jtruenary:
Since the last release [v0.36.7](https://github.com/codemagic-ci-cd/cli-tools/releases/tag/v0.36.7), the action get-latest-testflight-build-number returns version number instead of build number. Our Codemagic CI pipelines are breaking because of this. Any suggestions?

priitlatt:
Hi @jtruenary, thank you for your report. If anything, then version [`0.36.6`](https://github.com/codemagic-ci-cd/cli-tools/releases/tag/v0.36.6) is to blame for it as it includes changes from PR #300, that altered the behaviour of `app-store-connect get-latest-testflight-build-number` (and similar actions).

Could you please provide some more context of your use-case with terminal output and screenshots from the related application TestFlight builds?

mazzonem:
I'm having exactly the same issue.

It is generally believed that BuildNumber always goes up but Apple allows reusing (lower build number) when changing the app version. So Overall #300 is correct and we should adapt our CI pipelines @jtruenary.  https://developer.apple.com/library/archive/technotes/tn2420/_index.html#//apple_ref/doc/uid/DTS40016603-CH1-VERSION_NUMBER_AND_BUILD_NUMBER_CHECKLIST 

I hope it helps the next person looking for this 👍

jtruenary:
@priitlatt Yes you are right. Our last build was on 0.36.5 before this failing build, so our pipelines never ran on 0.36.6. 

We run this command: 

`app-store-connect get-latest-testflight-build-number "$APP_STORE_ID"`

This is the output:

`Found build number 1.18.35.104.6 from TestFlight version 6`

Previously we were fetching the build number and incrementing it in our build job,

priitlatt:
@jtruenary could you please provide a screenshot of what is going on in your TestFlight?

priitlatt:
Closing due to inactivity.

PR #304, opened 2023-01-23
App Store connect latest build number detection improvements

### LINKED ISSUE #305: Action get-latest-testflight-build-number returns number of expired build number instead of normal build number
Hello guys, 
Using the app-store-connect tool to get the latest testflight build number returns the build number of an expired build which isn't what was happening before the current version of the tool (0.36.7)
image
This is how our testflight looks:
image

priitlatt:
This is a follow up to PR #300.

The changes that were introduced in #300 expected that when [App Store](https://developer.apple.com/documentation/appstoreconnectapi/list_all_app_store_versions_for_an_app) or [prerelease](https://developer.apple.com/documentation/appstoreconnectapi/list_prerelease_versions) versions were listed for an app so that builds are also included in the responses, then there is a correspondence between the included resources (builds in our case) and the queried versions. However this does not always hold and as such the approach can lead to unexpected results.

Changes here choose the highest version (App Store or prerelease) that has builds and then returns the highest build number from this patch of builds.

In addition to this, add mutually exclusive optional arguments `--expired` and `--not-expired` to action `app-store-connect get-latest-testflight-build-number` to include or exclude expired builds from the latest build number lookup. (Fixes #305)

<details>
<summary>Example usage of new flags</summary>

```shell
$ app-store-connect get-latest-build-number "1481211155"
Found build number 2.0.320.508 from TestFlight version 2.0.320
2.0.320.508
$ app-store-connect get-latest-testflight-build-number "1481211155"
Found build number 2.0.320.508 from TestFlight version 2.0.320
2.0.320.508
$ app-store-connect get-latest-testflight-build-number "1481211155" --pre-release-version "01.04.01"
Found build number 134 from TestFlight version 01.04.01
134
$ app-store-connect get-latest-testflight-build-number "1481211155" --pre-release-version "01.04.01" --expired
Found build number 134 from TestFlight version 01.04.01
134
$ app-store-connect get-latest-testflight-build-number "1481211155" --pre-release-version "01.04.01" --not-expired
Did not find latest build for app 1481211155
```
</details>

**Updated actions**
- `app-store-connect get-latest-testflight-build-number`
- `app-store-connect get-latest-app-store-build-number`
- `app-store-connect get-latest-build-number`


artemii-yanushevskyi (approved):

artemii-yanushevskyi (approved):

nouman-hanif:
this change is breaking backward compatibility. I just spent half a day fixing the issue with my deployment pipeline. The default behavior without --expired or --not-expired switches was to return highest build number from builds that weren't expired. After this update, the behavior changed for the older deployment and it started returning highest build number from expired and non-expired combined. This should have remained unchanged. 

Also, the documentation of the script is not updated so there was no way to quickly figure out what had gone wrong. Can you please make sure to update the documentation? 

Thanks

priitlatt:
Hi @nouman-hanif. We're terribly sorry for the inconveniences that those changes have brought up to you and your team. 
The original (breaking) change came from PR #300 and its aftermath is discussed under #303. We were well aware that this change is going to break some workflows, but as the previous solution was downright wrong as build with version `2.0 (3)` is more recent than `1.9 (99)` by the rules that Apple has set (the latter would have been chosen before changes), then the behaviour of the action really had to be updated.

nouman-hanif:
@priitlatt thanks for the explanation. I understand sometimes to improve a solution it is unavoidable to maintain backward compatibility. The reason it took me more time to figure was because the documentation of the action on GitHub wasn't up to date with the changes. I am talking about readme at [get-latest-testflight-build-number](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/get-latest-testflight-build-number.md). If the two new optional switches` --expired --not-expired` can be added then anybody facing similar problem can figure out quickly. I ended up on this doc page from Codemagic docs topic [Automatic build versioning](https://docs.codemagic.io/knowledge-codemagic/build-versioning/), where I would imagine almost everybody would first look. 

Nouman

priitlatt:
@nouman-hanif Web docs for [`aapp-sotre-connect get-latest-testflight-build-number`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/get-latest-testflight-build-number.md) were updated in commit 
ce1fb8f and reflect `--expired` and `--not-expired` flags now.

nouman-hanif:
great!! @priitlatt :) thanks for the quick update. much appreciated!!

PR #308, opened 2023-02-06
Bugfix: Do not rely on provisioning profile ordering on disk for `xcode-project use-profiles`

priitlatt:
Two invocations of `xcode-project use-profiles` can lead to different modifications to Xcode project files even though the very same set of provisioning profiles are discovered. This can happen because the script [`code_signing_manager.rb`](https://github.com/codemagic-ci-cd/cli-tools/blob/v0.38.1/src/codemagic/scripts/code_signing_manager.rb), which is used internally for updating `*.xcodeproj` files, depends on the order of provisioning profiles that are passed to it.

In order to avoid such inconsistencies, make sure that profiles are sorted in a stable manner before serializing them for `code_signing_manager.rb`. The changes in this PR ensure that profiles are ordered so that
- Non-wildcard/strict matching provisioning profiles have higher priority than wildcard profiles.
- The more specific the bundle identifier is, the higher the priority. For example
  - `com.example.app.extension` is more specific than `com.example.app` and
  - `com.example.app.*` is more specific than `com.example.*`, which is more specific than just `*`.
- Provisioning profiles with the same level of namespacing (number of dots in bundle identifier) are order alphabetically by comparing bundle identifier values.
- Finally in case of bundle identifiers are the same, more recently generated provisioning profile comes first.

For example this is how those provisioning profiles would be ordered.

| Bundle identifier                       | Created at          |
|-----------------------------------------|---------------------|
| `io.codemagic.banaan.UITests.xctrunner` | 2023-01-20 13:17:03 |
| `io.codemagic.banaan.Clip`              | 2023-01-20 13:17:03 |
| `io.codemagic.banaan`                   | 2023-01-20 13:17:02 |
| `io.codemagic.capybara`                 | 2023-01-20 13:29:58 |
| `io.codemagic.*`                        | 2022-09-28 10:40:41 |
| `io.codemagic.*`                        | 2022-03-24 10:02:42 |
| `*`                                     | 2023-02-02 13:14:11 |
| `*`                                     | 2022-11-14 10:15:54 |

**Updated actions:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-Version 0.38.1
+Version 0.38.2
 -------------
 
 **Bugfix**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@ Version 0.38.1
 -------------
 
 **Bugfix**
+- Improve action `xcode-project use-profiles` stability so that different invocation with the same set  of provisioning profiles will always yield the same changeset to Xcode project settings. [PR #308](https://github.com/codemagic-ci-cd/cli-tools/pull/308)
+
+Version 0.38.1
+-------------
+
+**Bugfix**
 - Update flag `--cancel-previous-submissions` for the `app-store-connect builds submit-to-app-store` action to wait for Apple's confirmation that the submission is cancelled before attempting to submit a new build.
 
 Version 0.38.0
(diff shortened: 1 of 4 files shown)


VeArnold (approved):
:shipit:

VeArnold (approved):

PR #309, opened 2023-02-07
Dependencies: Remove upper bound from `cryptography` dependency

priitlatt:
Back when `codemagic-cli-tools` still had a direct dependency on both [`cryptography`](https://cryptography.io/en/latest/) and [`pyOpenSSL`](https://www.pyopenssl.org/en/latest/) libraries, then using `cryptography` version  `>= 38.0.0` caused 
```python
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'
```
arising from `OpenSSL.crypto` module.
Due to this an upper limit with constraint `<38.0.0` was enforced on `cryptography` in #268.

While direct dependency on `pyOpenSSL` was removed in #269 the constraint on `cryptography` version was never lifted.

Change in this PR removes upper bound from allowed `cryptography` versions making `codemagic-cli-tools` compatible with contemporary up-to-date releases of `cryptography`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ Version 0.39.0
 -------------
 
 **Dependencies**
-- Remove upper bound version limit from [`cryptography`](https://cryptography.io/) dependency. [PR #309](https://github.com/codemagic-ci-cd/cli-tools/pull/309)
+- Remove upper version limit from [`cryptography`](https://cryptography.io/) dependency. [PR #309](https://github.com/codemagic-ci-cd/cli-tools/pull/309)
 
 Version 0.38.3
 -------------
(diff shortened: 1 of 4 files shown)


priitlatt:
To be tested with:
- Python 3.8.x
  - cryptography 36.0.2
  - cryptography 37.0.1
  - cryptography 37.0.4
  - cryptography 38.0.0
  - cryptography 38.0.3
  - cryptography 39.0.0
- Python 3.11.x
  - cryptography 36.0.2
  - cryptography 37.0.1
  - cryptography 37.0.4
  - cryptography 38.0.0
  - cryptography 38.0.3
  - cryptography 39.0.0

PR #313, opened 2023-03-02
Use timezone aware datetimes

VeArnold:
As per Python `datetime` [documentation](https://docs.python.org/3/library/datetime.html), due to the fact that naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling `datetime.now(timezone.utc)`.

### DIFF
--- a/src/codemagic/utilities/auditing/http_request_auditor.py
+++ b/src/codemagic/utilities/auditing/http_request_auditor.py
@@ -1,6 +1,7 @@
 import re
 import urllib.parse
 from datetime import datetime
+from datetime import timezone
 from typing import Dict
 from typing import Optional
 
@@ -76,7 +77,7 @@ class HttpRequestAuditor(BaseAuditor):
                 'content': self._serialize_response_content(),
                 'elapsed': self._response.elapsed.total_seconds(),
             },
-            'timestamp': datetime.utcnow().isoformat(),
+            'timestamp': datetime.now(timezone.utc).isoformat(),
             'version': __version__,
         }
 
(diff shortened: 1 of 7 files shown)


src/codemagic/utilities/auditing/http_request_auditor.py
```diff
@@ -76,7 +77,7 @@ def _serialize_audit_info(self):
                 'content': self._serialize_response_content(),
                 'elapsed': self._response.elapsed.total_seconds(),
             },
-            'timestamp': datetime.utcnow().isoformat(),
+            'timestamp': datetime.now(timezone.utc).isoformat(),
```
artemii-yanushevskyi, line 80: Is it something to worry about?

```
>>> from datetime import timezone, datetime
>>> datetime.utcnow().isoformat()
'2023-03-02T11:47:16.262548'
>>> datetime.now(timezone.utc).isoformat()
'2023-03-02T11:47:20.887312+00:00'
```

VeArnold: Yeah, I thought about it too, but I thought it wouldn't make much of a difference, but I might be wrong... I will give it a more thorough thought soon.

Alternatively, we could keep `utcnow()` here, or ensure that we get the same output as before by using:
`datetime.now(timezone.utc).replace(tzinfo=None).isoformat()`

but that would kind of be weird and beat the purpose of it... I'll recheck soon.

priitlatt (approved):

src/codemagic/utilities/auditing/http_request_auditor.py
priitlatt: All good.

Issue #321, opened 2023-06-19
Firebase app distribution get latest build version is not working


Buddybetadmin:
We are trying to use the latest cli-tools firebase-app-distribution utility to fetch the latest build number from firebase using your documentations on codemagic website as well as github. 

```shell
$LATEST_BUILD_VERSION_ANDROID=$(firebase-app-distribution get-latest-build-version -p $FIREBASE_PROJECT_ID -a $Android_Firebase_App_ID_QA)
```

but we keep getting this weird error: 
`Client error. name: firebaseappdistribution  version: v1`

It seems like there is something wrong with the utility. Or we doing something wrong?

priitlatt:
Hi @Buddybetadmin! Thanks for bringing this up. The error you mentioned originates from too old Google API client library that the tools were using. This is fixed with [version 0.40.2](https://github.com/codemagic-ci-cd/cli-tools/releases/tag/v0.40.2).

PR #317, opened 2023-05-19
Feature: Firebase releases

artemii-yanushevskyi:
This update introduces a new `firebase-app-distribution` tool to list releases and retrieve the latest release build version from Firebase using Firebase API.

**New actions with documentation:**
- `firebase-app-distribution releases list`: list all releases associated with a Firebase application
- `firebase-app-distribution get-latest-build-version`: get the build version number for the latest release for a Firebase application

**New API methods**:
- `FirebaseAppDistribution.list_releases`
- `FirebaseAppDistribution.get_latest_build_version`

**Examples**:

<details>
<summary><code>firebase-app-distribution get-latest-build-version -p 228333310124 -a 1:228333310124:ios:5e439e0d0231a788ac8f09</code></summary>
<pre><code>
71
</code></pre>
</details>

<details>
<summary><code>firebase-app-distribution releases list -p 228333310124 -a 1:228333310124:ios:5e439e0d0231a788ac8f09 --json --limit 1 --order-by createTime</code></summary>
<pre><code>
[
    {
        "name": "projects/228333310124/apps/1:228333310124:ios:5e439e0d0231a788ac8f09/releases/0fam13pr3rea0",
        "releaseNotes": {
            "text": "Something new :) :)"
        },
        "displayVersion": "0.0.42",
        "buildVersion": 70,
        "createTime": "2023-05-18T12:30:17.454581+00:00",
        "firebaseConsoleUri": "https://console.firebase.google.com/project/capybara-application/appdistribution/app/ios:io.codemagic.capybara/releases/0fam13pr3rea0",
        "testingUri": "https://appdistribution.firebase.google.com/testerapps/1:228333310124:ios:5e439e0d0231a788ac8f09/releases/0fam13pr3rea0",
[...]

PR #322, opened 2023-06-21
Dependencies: Update minumum `google-api-python-client` version

priitlatt:
In #317 a new tool called `firebase-app-distribution` was added.

Using the [Firebase App Distribution releases APIs](https://firebase.google.com/docs/reference/app-distribution/rest/v1/projects.apps.releases) requires `google-api-python-client >= 2.84.0`, but up until this PR Codemagic CLI tools onlyonly required version `1.7.12` of `google-api-python-client`, which is incompatible with recent Google APIs.

Too old `google-api-python-client` version caused the invocations of `firebase-app-distribution` commands to fail with error
```
 Client error. name: firebaseappdistribution  version: v1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 0.40.2
+-------------
+
+**Dependencies**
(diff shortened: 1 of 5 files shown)


priitlatt:
Fixes #321.

PR #329, opened 2023-07-06
Feature: register devices from file

codemagic-mohammed:
- Introduce `app-store-connect register-devices-from-file` action to bulk-register devices from a file with UDIDs
- Add short flags for `app-store-connect register-device` action

<details>
<summary>Demonstration of the command and its output</summary>

## File content

```
00000000-000000000000001E
00000000-000000000000002E

00000000-000000000000003X
```

UDIDs in file:

1. Valid UDID, but the device is present. 
2. Valid UDID, the device doesn't yet exists and will be registered
3. Invalid UDID

## Command
```shell
app-store-connect register-devices-from-file -n iPhone UDIDs.txt
```

## Output
```
Creating new Device: udid: 00000000-000000000000001E, name: iPhone, platform: IOS
Failed to register a device: POST https://api.appstoreconnect.apple.com/v1/devices returned 409: There is a problem with the request entity - A device with number '00000000-000000000000001E' already exists on this team.

Creating new Device: udid: 00000000-000000000000002E, name: iPhone, platform: IOS
-- Device (Created) --
Id: H2985X3P8Q
Type: devices
Class: IPOD
Name: iPhone
Platform: IOS
Status: ENABLED
Udid: 00000000-000000000000002E
Created Device H2985X3P8Q

Invalid device UDID on line 4: '00000000-000000000000003X'
```

The syntax highlight demonstration
image

</details>


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "codemagic-cli-tools"
-version = "0.40.4"
+version = "0.40.5"
```
priitlatt, line 3: New feature should bump minor version, not only patch. The latter is for bugfixes and marginal improvements.

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -1215,17 +1215,21 @@ class DeviceArgument(cli.Argument):
     )
     DEVICE_NAME = cli.ArgumentProperties(
         key='device_name',
-        flags=('--name',),
+        flags=('-n', '--name'),
```
priitlatt, line 1218:
```suggestion
        flags=('--name', '-n'),
```

Shorthand flag is expected to come after the long version. Then both command line and markdown docs refer to the descriptive name before the more cryptic shorthand. (Note the changes in autogenerated docs).

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -1215,17 +1215,21 @@ class DeviceArgument(cli.Argument):
...
         argparse_kwargs={'required': False},
     )
+    COMMON_DEVICE_NAME = cli.ArgumentProperties.duplicate(
+        DEVICE_NAME,
+        description='Common name of the Devices',
+    )
     DEVICE_UDID = cli.ArgumentProperties(
         key='device_udid',
-        flags=('--udid',),
+        flags=('-u', '--udid'),
```
priitlatt, line 1232:
```suggestion
        flags=('--udid', '-u'),
```
Same as above.

src/codemagic/tools/app_store_connect.py
```diff
@@ -1195,6 +1242,10 @@ def _save_certificates(
     ) -> List[pathlib.Path]:
         return [self._save_certificate(c, private_key, p12_container_password) for c in certificates]
 
+    @staticmethod
+    def _is_device_udid_valid(device_udid: str) -> bool:
```
priitlatt, line 1246: I find that this method would be better suited under `Device` class. Also, could you provide a documentation reference for this UDID structure?

src/codemagic/tools/app_store_connect.py
```diff
@@ -510,6 +513,50 @@ def register_device(
...
+        DeviceArgument.COMMON_DEVICE_NAME,
+        DeviceArgument.DEVICE_UDIDS_PATH,
+    )
+    def register_devices_from_file(
+        self,
+        platform: BundleIdPlatform,
+        device_name: str,
+        device_udids_path: pathlib.Path,
+        should_print: bool = True,
+    ) -> List[Device]:
```
priitlatt, line 528: This action declaration / definition contradicts pretty much all existing principles we've had. There is no other action which reads input only from file and here too it would be expected that the user can register multiple devices using a single command by just passing multiple UDIDs instead of one. Reading from file (and environment variable) can be achieved with custom environment argument, which also does the validation.

priitlatt (changes requested):
I don't like the proposed approach and feel like it would make a lot more sense to boost the functionality of the already existing `register-device` action.

src/codemagic/tools/app_store_connect.py
artemii-yanushevskyi: I thought to do just the way you proposed (add support for file to existing action and allow to pass multiple values). The problem was that the existing action raises an error for duplicate devices, while this action should ignore duplicates and move on

This would cause a major with a call like

```sh
app-store-connect register-device -u 00000000-000000000000000E -n iPhone --udids-path UDIDs.txt
```

This is the reason I created a separate action that would only accept UDIDs from a file, hence the `from-file` suffix.

As I had more time to think about it, I believe we can resolve this inconsistency by introducing a strict mode and have it enabled by default (otherwise the existing `register-device` would stop returning 1 exit code on error, it would be breaking change)

src/codemagic/tools/app_store_connect.py
artemii-yanushevskyi: Sorry, I inferred this UDID structure from the existing values, so no documentation links.
Should we remove this validation?

src/codemagic/tools/app_store_connect.py
priitlatt: As long as the default behaviour remains the same as it used to then we're good. Adding a boolean switch would work fine for that purpose.

src/codemagic/tools/app_store_connect.py
priitlatt: This is not an official source, but from [here](https://getupdraft.com/get-my-udid) I can see that
> There are **two formats of ID**: iOS devices that came on the market between 2007 and 2018 have a 40-digit lowercase hexadecimal code and all devices after 2018 have a 25-digit uppercase hexadecimal code.

Perhaps we can omit validation on our end and just require the UDID not to be empty string, and leave all verification to Apple.

artemii-yanushevskyi:
Declining in favor of https://github.com/codemagic-ci-cd/cli-tools/pull/331.

PR #331, opened 2023-07-11
Feature: register-device supports multiple UDIDs

artemii-yanushevskyi:
1. Update `app-store-connect register-device` to accept now accepts multiple UDIDs from a file or an environment variable as a source.
2. Introduce `--ignore-registration-errors` flag for `app-store-connect register-device` to continue registering devices despite errors.
3. Introduce short flags for `app-store-connect register-device`:
    - `-n` flag for the device name,
    - `-u` flag for the UDIDs.

<details>
<summary>Demonstration of the command and its output</summary>

## File content

```
00000000-000000000000001E
00000000-000000000000002E

00000000-000000000000003X
```

UDIDs in file:

1. Valid UDID, but the device is present. 
2. Valid UDID, the device doesn't yet exists and will be registered
3. Invalid UDID

## Command
```shell
app-store-connect register-device -n iPhone -u @file:Device\ UDIDs.txt --ignore-registration-errors
```

## Output
```
Creating new Device: udid: 00000000-000000000000001E, name: iPhone, platform: IOS
Failed to register a device: POST https://api.appstoreconnect.apple.com/v1/devices returned 409: There is a problem with the request entity - A device with number '00000000-000000000000001E' already exists on this team.

Creating new Device: udid: 00000000-000000000000002E, name: iPhone, platform: IOS
-- Device (Created) --
Id: H2985X3P8Q
Type: devices
Class: IPOD
Name: iPhone
Platform: IOS
Status: ENABLED
Udid: 00000000-000000000000002E
Created Device H2985X3P8Q

Creating new Device: udid: 00000000-000000000000003X, name: iPhone, platform: IOS
Failed to register a device: POST https://api.appstoreconnect.apple.com/v1/devices returned 409: An attribute in the provided entity has invalid value - An invalid value '00000000-000000000000003X' was provided for the parameter 'udid'.
```

The syntax highlight demonstration
image

</details>

Previous [PR](https://github.com/codemagic-ci-cd/cli-tools/pull/329) for the related functionality.

### DIFF
--- a/poetry.toml
+++ b/poetry.toml
@@ -1,3 +1,0 @@
-[virtualenvs]
-create = true
-in-project = true
--- a/poetry.toml
+++ b/poetry.toml
@@ -1,0 +1,3 @@
+[virtualenvs]
+create = true
+in-project = true
--- a/src/codemagic/tools/_app_store_connect/arguments.py
+++ b/src/codemagic/tools/_app_store_connect/arguments.py
@@ -364,9 +364,9 @@ class Types:
                 udids = [non_typed_value.strip()]
             else:
                 udids = [udid.strip() for udid in shlex.split(non_typed_value) if udid.strip()]
-            if not udids or not all(udids):
-                raise argparse.ArgumentTypeError(f'Provided value "{non_typed_value}" is not valid')
-            return udids
+            if udids and all(udids):
+                return udids
+            raise argparse.ArgumentTypeError(f'Provided value "{non_typed_value}" is not valid')
 
 
 _API_DOCS_REFERENCE = f"Learn more at {AppStoreConnectApiClient.API_KEYS_DOCS_URL}."
--- a/src/codemagic/tools/_app_store_connect/arguments.py
+++ b/src/codemagic/tools/_app_store_connect/arguments.py
@@ -359,7 +359,11 @@ class Types:
         example_value = "00000000-000000000000001E"
 
         def _apply_type(self, non_typed_value: str) -> List[str]:
-            udids = [udid.strip() for udid in shlex.split(non_typed_value) if udid.strip()]
+            is_from_cli = not (self._is_from_environment() or self._is_from_file() or self._from_environment)
+            if is_from_cli:
+                udids = [non_typed_value]
+            else:
+                udids = [udid.strip() for udid in shlex.split(non_typed_value) if udid.strip()]
             if not udids or not all(udids):
                 raise argparse.ArgumentTypeError(f'Provided value "{non_typed_value}" is not valid')
             return udids
@@ -1228,7 +1232,7 @@ class DeviceArgument(cli.Argument):
         argparse_kwargs={"required": False},
     )
     DEVICE_UDIDS = cli.ArgumentProperties(
-        key="device_udids_argument",
+        key="device_udids",
         flags=("--udid", "-u"),
         type=Types.DeviceUdidsArgument,
         description=f"Device ID (UDID), for example: {Types.DeviceUdidsArgument.example_value}",
--- a/src/codemagic/tools/_app_store_connect/arguments.py
+++ b/src/codemagic/tools/_app_store_connect/arguments.py
@@ -355,18 +355,13 @@ class Types:
 
     class DeviceUdidsArgument(cli.EnvironmentArgumentValue[List[str]]):
         argument_type = List[str]
-        environment_variable_key = "DEVICE_UDIDS"
+        environment_variable_key = "APP_STORE_CONNECT_DEVICE_UDIDS"
         example_value = "00000000-000000000000001E"
 
         def _apply_type(self, non_typed_value: str) -> List[str]:
-            if self._is_from_environment() or self._is_from_file():
-                udids = [udid.strip() for udid in shlex.split(non_typed_value)]
-            else:
-                udids = [non_typed_value.strip()]
-
+            udids = [udid.strip() for udid in shlex.split(non_typed_value) if udid.strip()]
             if not udids or not all(udids):
                 raise argparse.ArgumentTypeError(f'Provided value "{non_typed_value}" is not valid')
-
             return udids
 
 
@@ -1233,7 +1228,7 @@ class DeviceArgument(cli.Argument):
         argparse_kwargs={"required": False},
     )
     DEVICE_UDIDS = cli.ArgumentProperties(
-        key="device_udids",
+        key="device_udids_argument",
         flags=("--udid", "-u"),
         type=Types.DeviceUdidsArgument,
         description=f"Device ID (UDID), for example: {Types.DeviceUdidsArgument.example_value}",
--- a/src/codemagic/tools/_app_store_connect/arguments.py
+++ b/src/codemagic/tools/_app_store_connect/arguments.py
@@ -361,7 +361,7 @@ class Types:
         def _apply_type(self, non_typed_value: str) -> List[str]:
             is_from_cli = not (self._is_from_environment() or self._is_from_file() or self._from_environment)
             if is_from_cli:
-                udids = [non_typed_value]
+                udids = [non_typed_value.strip()]
             else:
                 udids = [udid.strip() for udid in shlex.split(non_typed_value) if udid.strip()]
             if not udids or not all(udids):
@@ -1224,7 +1224,7 @@ class DeviceArgument(cli.Argument):
     DEVICE_NAME = cli.ArgumentProperties(
         key="device_name",
         flags=("--name", "-n"),
-        description="Name of the Device",
+        description="Common name of Devices",
         argparse_kwargs={"required": True},
     )
     DEVICE_NAME_OPTIONAL = cli.ArgumentProperties.duplicate(
@@ -1260,7 +1260,11 @@ class DeviceArgument(cli.Argument):
             "Ignore device registration failures, e.g. invalid UDID or duplicate UDID submission. "
             "Proceed registering remaining UDIDs when the flag is set."
         ),
-        argparse_kwargs={"required": False, "action": "store_true"},
+        argparse_kwargs={
+            "required": False,
+            "action": "store_true",
+            "default": False,
+        },
     )
 
 
--- a/src/codemagic/tools/_app_store_connect/arguments.py
+++ b/src/codemagic/tools/_app_store_connect/arguments.py
@@ -357,6 +357,17 @@ class Types:
 
             return beta_build_infos
 
+    class DeviceUdidsArgument(cli.EnvironmentArgumentValue[List[str]]):
+        environment_variable_key = "DEVICE_UDIDS"
+        example_value = "00000000-000000000000001E"
+
+        @classmethod
+        def _apply_type(cls, non_typed_value: str) -> List[str]:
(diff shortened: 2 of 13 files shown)


src/codemagic/tools/app_store_connect.py
```diff
@@ -492,25 +492,41 @@ def list_devices(
...
-        DeviceArgument.DEVICE_UDID,
+        DeviceArgument.DEVICE_UDIDS,
+        DeviceArgument.IGNORE_REGISTRATION_ERRORS,
     )
     def register_device(
         self,
         platform: BundleIdPlatform,
         device_name: str,
-        device_udid: str,
+        device_udids: Types.DeviceUdidsArgument,
```
priitlatt, line 502: That breaks backwards compatibility.

src/codemagic/tools/app_store_connect.py
```diff
@@ -492,25 +492,41 @@ def list_devices(
...
+                device = self._create_resource(
+                    self.api_client.devices,
+                    should_print,
+                    udid=device_udid,
+                    name=device_name,
+                    platform=platform,
+                )
+            except AppStoreConnectError as error:
+                if not ignore_registration_errors:
+                    raise error from error
```
priitlatt, line 524: Use just `raise` as doing `raise error from error` causes unnecessary traceback stacking.

<details>
<summary>Example</summary>

```python
try:
    1/0
except ZeroDivisionError as zde:
    print("Oopsie, division by zero")
    raise zde from zde
```
runs as
```python
Oopsie, division by zero
  File "/private/tmp/error.py", line 5, in <module>
    raise zde from zde
  File "/private/tmp/error.py", line 2, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero
```
whereas
```python
try:
    1/0
except ZeroDivisionError as zde:
    print("Oopsie, division by zero")
    raise
```
points directly to the original cause
```python
Oopsie, division by zero
  File "/private/tmp/error.py", line 2, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero
```

</details>

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -1212,18 +1223,19 @@ class DeviceArgument(cli.Argument):
...
-    DEVICE_UDID = cli.ArgumentProperties(
-        key="device_udid",
-        flags=("--udid",),
-        description="Device ID (UDID)",
+    DEVICE_UDIDS = cli.ArgumentProperties(
+        key="device_udids",
+        flags=("--udid", "-u"),
+        type=Types.DeviceUdidsArgument,
+        description=f"Device ID (UDID), for example: {Types.DeviceUdidsArgument.example_value}",
         argparse_kwargs={"required": True},
```
priitlatt, line 1239: It cannot be required argument if we want to allow environment variable overriding.

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -357,6 +357,17 @@ def _apply_type(cls, non_typed_value: str) -> List[BetaBuildInfo]:
 
             return beta_build_infos
 
+    class DeviceUdidsArgument(cli.EnvironmentArgumentValue[List[str]]):
+        environment_variable_key = "DEVICE_UDIDS"
+        example_value = "00000000-000000000000001E"
+
+        @classmethod
+        def _apply_type(cls, non_typed_value: str) -> List[str]:
+            udids = non_typed_value.split()
```
priitlatt, line 366: `shlex.split` would be safer to use. Better yet, filter out empty strings after stripping them.

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -357,6 +357,17 @@ def _apply_type(cls, non_typed_value: str) -> List[BetaBuildInfo]:
...
+            if not udids:
```
priitlatt, line 367: Depending on the way how `udids` list is constructed that check may not be sufficient. Consider input such as `'"" " "'`. Then from shell perspective the passed arguments are `['', ' ']` (empty string and just whitespace). We can be (quite) sure that neither are valid UDIDs.

src/codemagic/tools/_app_store_connect/arguments.py
priitlatt, line 1239: If we don't allow multiple args on argparse level, then documentation does not indicate that multiple args are allowed. This applies to both MD and CLI docs.

priitlatt (changes requested):
Unfortunately the proposed approach does not work. On app level device UDIDs argument is not treated in a way that is would properly support multiple arguments. Consequently it yields very mixed behaviour:
```shell
app-store-connect register-device -u 1 2 3 4 -n "device-name"
```
raises an error 
```
error: unrecognized arguments: 2 3 4
```
while very unexpectedly (as per docs)
```shell
app-store-connect register-device -u "1 2 3 4" -n "device-name" --ignore-registration-errors
```
works by trying to add 4 devices.


Passing values using environment variable does not work either:
```shell
DEVICE_UDIDS="1 2" app-store-connect register-device -n "device-name"
```
as it fails with 
```
app-store-connect register-device: error: the following arguments are required: --udid/-u
```

src/codemagic/tools/app_store_connect.py
```diff
@@ -492,25 +492,53 @@ def list_devices(
...
-        DeviceArgument.DEVICE_UDID,
+        DeviceArgument.DEVICE_UDIDS,
+        DeviceArgument.IGNORE_REGISTRATION_ERRORS,
     )
     def register_device(
         self,
         platform: BundleIdPlatform,
         device_name: str,
-        device_udid: str,
+        device_udids_argument: Union[str, Types.DeviceUdidsArgument, Sequence[Union[str, Types.DeviceUdidsArgument]]],
```
priitlatt, line 502: Given that `DeviceArgument.DEVICE_UDIDS` is not required, then the action implementation should accept `None` values too and raise an error if the value is missing. Currently execution fails with an unexpected exception if no UDIDs are specified.

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -357,6 +353,17 @@ def _apply_type(cls, non_typed_value: str) -> List[BetaBuildInfo]:
...
+    class DeviceUdidsArgument(cli.EnvironmentArgumentValue[List[str]]):
+        argument_type = List[str]
+        environment_variable_key = "APP_STORE_CONNECT_DEVICE_UDIDS"
+        example_value = "00000000-000000000000001E"
+
+        def _apply_type(self, non_typed_value: str) -> List[str]:
+            udids = [udid.strip() for udid in shlex.split(non_typed_value) if udid.strip()]
+            if not udids or not all(udids):
+                raise argparse.ArgumentTypeError(f'Provided value "{non_typed_value}" is not valid')
+            return udids
```
priitlatt, line 365: Splitting should be applied only for arguments that originate from file or environment variable. 
Consider this:
```shell
app-store-connect register-device -u "first udid" "another udid" -n "device-name" --ignore-registration-errors
```
I'd expect that this invocation attempts to register two UDIDs: `first udid` and `another udid`. In practice it attempts to register four UDIDs: `first`, `udid`, `another` and again `udid`. 

Note that when providing the values via environment variable, it is all good.
```shell
APP_STORE_CONNECT_DEVICE_UDIDS="udid1 udid2 'udid 3'" app-store-connect register-device -n "device-name" --ignore-registration-errors
```
Running the above indeed attempts to register three UDIDs: `udid1`, `udid2` and `udid 3` (note the whitespace).

To overcome this you should check the origin of the argument:
```python
is_from_cli = not (self._is_from_environment() or self._is_from_file() or self._from_environment)
```
and apply splitting only in cases if the argument is not provided directly via CLI.

Note that as of now the above check will cause an error as of now since `_from_environment` is defined as the last thing in the `__init__` method of `TypedCliArgument`. This can be easily fixed though.

src/codemagic/tools/_app_store_connect/arguments.py
```diff
@@ -1212,19 +1219,24 @@ class DeviceArgument(cli.Argument):
...
         DEVICE_NAME,
         argparse_kwargs={"required": False},
     )
-    DEVICE_UDID = cli.ArgumentProperties(
-        key="device_udid",
-        flags=("--udid",),
-        description="Device ID (UDID)",
-        argparse_kwargs={"required": True},
+    DEVICE_UDIDS = cli.ArgumentProperties(
+        key="device_udids_argument",
```
priitlatt, line 1231: Nowhere else we have `argument` suffix in the key name. Also this makes the Python API of actions rather ugly, especially when used with kwargs.

src/codemagic/tools/app_store_connect.py
```diff
@@ -492,25 +492,53 @@ def list_devices(
...
         DeviceArgument.DEVICE_NAME,
-        DeviceArgument.DEVICE_UDID,
+        DeviceArgument.DEVICE_UDIDS,
+        DeviceArgument.IGNORE_REGISTRATION_ERRORS,
     )
     def register_device(
         self,
         platform: BundleIdPlatform,
         device_name: str,
-        device_udid: str,
```
priitlatt, line 501: Please retain the old `device_udid: str` positional argument for backwards compatibility as optional on the same position.
```python
device_udid: Optional[str] = None,
```
We don't want to introduce breaking changes to Python API. Final UDIDs can be composed by combining new UDIDs args and the legacy one. If no UDIDs are given, then fail with error.

src/codemagic/tools/app_store_connect.py
```diff
@@ -492,25 +492,53 @@ def list_devices(
...
+        Register new Devices for app development
         """
-        return self._create_resource(
-            self.api_client.devices,
-            should_print,
-            udid=device_udid,
-            name=device_name,
-            platform=platform,
-        )
+        if isinstance(device_udids_argument, str) or isinstance(device_udids_argument, Types.DeviceUdidsArgument):
```
priitlatt, line 509: It should be checked that `device_udids_argument` is not `None`, which it can be as `DeviceArgument.DEVICE_UDIDS` is optional.

src/codemagic/tools/app_store_connect.py
```diff
@@ -492,25 +492,53 @@ def list_devices(
...
+                device_udids.extend(device_udids_argument.value)
+            else:
+                device_udids.append(device_udids_argument)
+        if not device_udids:
+            DeviceArgument.DEVICE_UDIDS.raise_argument_error("At least one device UDID is required")
+
+        registered_devices = []
+        for i, device_udids_argument in enumerate(device_udids):
+            if should_print:
+                self.echo("") if i != 0 else None
```
priitlatt, line 524: I find that it would be easier to remove the enumeration and do this echoing after the try-except-else block. Empty line after the last registration is fine.

priitlatt (changes requested):

poetry.toml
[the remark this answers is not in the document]
artemii-yanushevskyi: Sorry, I accidentally committed this file

priitlatt (changes requested):

priitlatt (approved):
All looks good now 💯

PR #340, opened 2023-08-26
Add support for publishing macOS builds to TestFlight and App Store

nilsreichardt:
Closes #339

Action [`app-store-connect publish`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/publish.md) can be used to upload iOS and macOS application packages (`ipa` and `pkg` files respectively) to App Store Connect. At the time of initial implementation, TestFlight beta distribution was not supported for macOS applications and as such using [`--testflight`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/publish.md#--testflight--t) option when publishing macOS packages is just ignored and has no effect.

Changes proposed here remove this limitation as TestFlight beta distribution for macOS apps has been supported for quite some time already. However, as Apple's processing procedure of macOS packages differs slightly from iOS packages, it became necessary to also modify the method that waits until build processing completes. For iOS it will suffice to just wait until [build's `processingState`](https://developer.apple.com/documentation/appstoreconnectapi/build/attributes) attribute is not `PROCESSING` any more. But behaviour for macOS apps is slightly different as [beta details](https://developer.apple.com/documentation/appstoreconnectapi/buildbetadetail) associated with the build might still be processed after that point. Hence `BuildsActionGroup.wait_until_build_is_processed` was updated to accommodate that.

Necessary App Store Connect API client method and object mappings for dealing with [BuildBetaDetail](https://developer.apple.com/documentation/appstoreconnectapi/buildbetadetail) were added along with new action to show beta detail info for given build:

<details>
<summary><code>app-store-connect builds beta-details</code> usage</summary>

```shell
$ app-store-connect builds beta-details "3bf3e846-3d31-4e1e-ab0f-0834fb9f9a26"
Get Build Beta Details for Build 3bf3e846-3d31-4e1e-ab0f-0834fb9f9a26
-- Build Beta Detail --
Id: 3bf3e846-3d31-4e1e-ab0f-0834fb9f9a26
Type: buildBetaDetails
Auto notify enabled: True
External build state: EXPIRED
Internal build state: EXPIRED
```
</details>

**Updated actions**
- `app-store-connect publish`

**New actions**
- `app-store-connect builds beta-details`

PR #343, opened 2023-09-07
Fix App Store Connect build processing waiting for macOS packages

priitlatt:
Prerequisite for #340.


priitlatt:
Everything from here is already included in #343.

PR #353, opened 2023-09-20
Add note that `--cancel-previous-submissions` does not work for TestFlight submissions

nilsreichardt:
From what I understand, is it not possible to cancel TestFlight submissions, right? I found the question "[Delete Beta App Review Submission via App Store Connect API](https://developer.apple.com/forums/thread/680143)" in the apple forum which asks this question. Still today, I can't find a delete option in the [API docs](https://developer.apple.com/documentation/appstoreconnectapi/prerelease_versions_and_beta_testers/beta_app_review_submissions).

When reading the docs for `app-store-connect publish` I thought it would also delete the previous TestFlight submissions when using the `--testflight` flag. I tried it out, but it didn't work (see [logs](https://github.com/SharezoneApp/sharezone-app/actions/runs/6248488396/job/16963390721)). Therefore, I clarified the docs about `--cancel-previous-submissions`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Version UNRELEASED
+-------------
+
+**Docs**
+- Update option `--cancel-previous-submissions` documentation for actions `app-store-connect builds submit-to-app-store` and `app-store-connect publish`. [PR #353](https://github.com/codemagic-ci-cd/cli-tools/pull/353)
+
+
 Version 0.45.3
 -------------
 
--- a/docs/app-store-connect/builds/submit-to-app-store.md
+++ b/docs/app-store-connect/builds/submit-to-app-store.md
@@ -50,7 +50,7 @@ Maximum amount of minutes to wait for the freshly uploaded build to be processed
 ##### `--cancel-previous-submissions`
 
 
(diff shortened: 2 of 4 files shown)


priitlatt (changes requested):
This option is indeed only applicable for App Store submissions and the documentation can be misleading as it is now.

That being said, markdown docs are autogenerated and not meant to be manually edited.
Instead descriptions of CLI options should be updated in code as those are the true sources of documentation. Then the changes are also reflected in command line `--help` messages.

Later on docs can be regenerated by running the following command from repository root:
```shell
poetry run python doc.py
```

In this particular case, the description of `--cancel-previous-submissions` flag should be updated in [`src/codemagic/tools/_app_store_connect/arguments.py` line 933](https://github.com/codemagic-ci-cd/cli-tools/blob/dbc5fcc8930145892723aae4b1f4ed3c575e6421/src/codemagic/tools/_app_store_connect/arguments.py#L933).

nilsreichardt:
Ahh, makes totally sense. I already wondered to keep all the files up to date. Thanks for the detailed instructions. I updated the PR 👍

priitlatt (approved):
All good now!

PR #364, opened 2023-10-24
Improve general error message for unexpected failures

priitlatt:
If action terminates unexpectedly with an error, then the error message that is shown to users has several issues.

Consider the following error for example:
```shell
$ google-play  get-latest-build-number --package-name "${UNDEFINED_PACKAGE_NAME}" --tracks "${UNDEFINED_TRACK_NAME}"                                                                        9 ↵
Get package "None" latest build number from track ""
Executing GooglePlay action get-latest-build-number failed unexpectedly. Detailed logs are available at "/var/folders/wr/c44p23x10f302_kfbj32z0p80000gn/T/codemagic-24-10-23.log". To see more details about the error, add `--verbose` command line option.
```

Here we can see that:
1. The class name of the tool is shown instead of the tool's executable name, which is little to no use for users.
2. If the invoked action is wrapped into action group, such as `app-store-connect apps get`, then the action group portion if completely omitted from log output,
3. Using `--verbose` does not actually show more details about the error in log output.

This PR aims to resolve the issues listed above:
1. Executable name is displayed in the error message.
2. Full executed command (without argument) is shown in the log output.
3. If `--verbose` flag is used, then exception stacktrace is included in stdout log output.

All in all the formatted error looks as follows:
```shell
$ app-store-connect apps get invalid-app-id --verbose
[15:24:14] WARNING > Executing "app-store-connect apps get" failed unexpectedly. Detailed logs are available at "/var/folders/wr/c44p23x10f302_kfbj32z0p80000gn/T/codemagic-24-10-23.log".
[15:24:14] ERROR > Exception traceback:
  File "/Users/priit/development/nevercode/cli-tools/src/codemagic/cli/cli_app.py", line 213, in invoke_cli
    CliApp._running_app._invoke_action(args)
  File "/Users/priit/development/nevercode/cli-tools/src/codemagic/cli/cli_app.py", line 170, in _invoke_action
    return cli_action(**action_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/priit/development/nevercode/cli-tools/src/codemagic/cli/cli_app.py", line 465, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/priit/development/nevercode/cli-tools/src/codemagic/tools/_app_store_connect/action_groups/apps_action_group.py", line 39, in get_app
    _ = 1 / 0
        ~~^~~
ZeroDivisionError: division by zero
```

Or without `--verbose` logging:
```shell
$ app-store-connect apps get invalid-app-id
Executing "app-store-connect apps get" failed unexpectedly. Detailed logs are available at "/var/folders/wr/c44p23x10f302_kfbj32z0p80000gn/T/codemagic-24-10-23.log". To see more details about the error, add "--verbose" command line option.
```


src/codemagic/cli/cli_app.py
```diff
@@ -104,16 +104,23 @@ def _parent_class_kwargs(cls, cli_args: argparse.Namespace) -> Dict[str, Any]:
...
+            f'{"" if args.verbose else verbose_log_suggestion}'
         )
-        logger.warning(Colors.RED(message))
-        file_logger = log.get_file_logger(cls)
-        file_logger.exception("Exception traceback:")
+
+        log.get_logger(cls).warning(Colors.RED(message))
+        logger = log.get_logger(cls, log_to_file=True, log_to_stream=args.verbose)
+        logger.exception("Exception traceback:")
         auditing.save_exception_audit()
```
artemii-yanushevskyi, line 124: Worth perhaps move
```py
logger.exception("Exception traceback:")
```

inside `auditing.save_exception_audit()`

And maybe even
```py
logger = log.get_logger(cls, log_to_file=True, log_to_stream=args.verbose)
```

artemii-yanushevskyi (changes requested):

src/codemagic/cli/cli_app.py
priitlatt: I don't agree with this. Currently exception logging and the preceding message are somewhat bound together:
- if verbose output is enabled, then part of the previous message is omitted and stdout stream logging is enabled,
- or if verbose output is disabled, then stdout logging is turned off and there is a statement that more details can be seen if `--verbose` is used.
In case `logger.exception` statement is moved to somewhere else, then this logic becomes disconnected and harder to maintain.

PR #394, opened 2024-03-14
Interrupt `altool`'s CFNetworking retry loop 

priitlatt:
Sometimes `app-store-connect publish` action gets stuck while binary is being uploaded using `altool`. From verbose `altool` logs it can be seen that uploading fails because specified hostname could not be found, and then internally `altool` just keeps on retrying the operation indefinitely.

<details>
<summary>Example of <code>altool</code> verbose log output</summary>

```log
2024-03-13 14:26:15.227 DEBUG: [ContentDelivery.Uploader] Wrote part 1 to temp file '/var/folders/w2/rrf5p87d1bbfyphxc7jdnyvh0000gn/T/com.apple.cds.vbtx/com.apple.cds_6A4F9402-15C4-49DA-8C2C-57950EFCD52F/Part-1.tmp'.
2024-03-13 14:26:15.227 DEBUG: [ContentDelivery.Uploader] Adding upload task for part 1.
2024-03-13 14:26:15.228 DEBUG: [ContentDelivery.Uploader] Upload tasks created. Waiting for them to finish uploading.
2024-03-13 14:26:15.236 DEBUG: [ContentDelivery.Uploader] LOST 0 bytes for part 1.
2024-03-13 14:26:15.236 DEBUG: [ContentDelivery.Uploader] Adding upload task 3 for part 1.
2024-03-13 14:26:15.236 *** Error: ERROR: [ContentDelivery.Uploader] WILL RETRY PART 1. Failed with error: Error Domain=NSURLErrorDomain Code=-1003 "A server with the specified hostname could not be found." UserInfo={_kCFStreamErrorCodeKey=8, NSUnderlyingError=0x6000039494a0 {Error Domain=kCFErrorDomainCFNetwork Code=-1003 "(null)" UserInfo={_kCFStreamErrorDomainKey=12, _kCFStreamErrorCodeKey=8, _NSURLErrorNWResolutionReportKey=Resolved 0 endpoints in 6ms using unknown from cache, _NSURLErrorNWPathKey=satisfied (Path is satisfied), interface: en0, ipv4, dns}}, _NSURLErrorFailingURLSessionTaskErrorKey=LocalUploadTask <89821E5C-11AB-4F69-B1CB-69AAC3D6FBB8>.<2>, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalUploadTask <89821E5C-11AB-4F69-B1CB-69AAC3D6FBB8>.<2>"
[...]

PR #395, opened 2024-03-18
Revert PR #394

priitlatt:
The changes applied in PR #394 were not effective and as such there is no point in persisting them since it adds both unnecessary complexity in code and considerable amount of runtime overhead.

Diff between `master` and this branch before #394 was merged: https://github.com/codemagic-ci-cd/cli-tools/compare/040b4abc8286c49faa74a58f32700c6e5377c8f9...f9c8666c35601974e6f988902f42d1da36c45849.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+Version 0.50.5
+-------------
+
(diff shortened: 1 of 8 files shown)


remarkov (approved):
special thanks for diff link, super useful for such cases, I will do the same next time I revert something 👍

PR #411, opened 2024-07-01
Fix exporting code signing certificates with OpenSSL 3.2.+

priitlatt:
It is not possible to add certifiacte and key to macOS keychain in PKCS#12 if the certificate subject contains unicode characters. OpenSLL versions prior to 3.2 used to encode the subject line in bag attributes such that no unicode literals were used, but this has been changed. As a result the following workflow is possibly broken:
```shell
app-store-connect certificates get <certificate-id> --save
keychain initialize
keychain add-certifiactes
```
as the last command can fail with error
```
security: SecKeychainItemImport: Unknown format in import.
```
in case the certificate subject contains non-ascii characters.

To overcome that, update PKCS#12 container creation workflow not to rely on `openssl` directly when creating the encrypted container, and instead utilize the tools that are built in to [`cryptography`](https://cryptography.io/en/latest/) library. Decryption (password removal from encrypted PKCS#12) is still dependent on `openssl` as the non-encrypted container created by `cryptography` uses `SHA256` as HMAC, which macOS Keychain does not understand. As a last step, the optional subject line is removed from the non-encrypted container manually to ensure that no non-ascii characters are present (subject under certificate's bag attributes is a mere convenience attribute as the same information is encoded in the certificate body anyway).

**Updated actions**
- `app-store-connect certificates get`
- `app-store-connect certificates list`
- `app-store-connect fetch-signing-files`


pyproject.toml
```diff
@@ -36,7 +36,7 @@ xcode-project = "codemagic.tools:XcodeProject.invoke_cli"
 
 [tool.poetry.dependencies]
 python = "^3.7"
-cryptography = ">= 3.3, != 37.0.0"
+cryptography = ">= 38.0.0"
```
priitlatt, line 39: Due to changes below we're now using [`PrivateFormat.PKCS12`](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.PrivateFormat.PKCS12) from `cryptography`, which was added only in version [38.0.0](https://pypi.org/project/cryptography/38.0.0/). But it should be fine as this version is already close to two years old and installing anything older than 35.0.0 is very questionable anyway.

priitlatt:
Tested on Codemagic stack:
- [x] Apple silicon Mac mini / Xcode 15.0
- [x] Apple silicon Mac mini / Xcode 15.1
- [x] Apple silicon Mac mini / Xcode 15.2
- [x] Apple silicon Mac mini / Xcode 15.3
- [x] Apple silicon Mac mini / Xcode 15.4
- [x] Apple silicon Mac mini / Xcode 16.0
- [x] Intel Mac mini / Xcode 14.2
- [x] Intel Mac mini / Xcode 14.1
- [x] Intel Mac mini / Xcode 13.2
- [x] Intel Mac mini / Xcode 12.5

pyproject.toml
artemii-yanushevskyi: Worth adding a note to changeling, perhaps?

priitlatt: Yes, of course! I started to rush a little last night.

priitlatt: @artemii-yanushevskyi, I updated changelog in [76dcb58](https://github.com/codemagic-ci-cd/cli-tools/pull/411/commits/76dcb58b9e2e26f863f89422803c2673d7ababa2). Will this suffice?

src/codemagic/models/certificate_p12_exporter.py
```diff
@@ -42,17 +87,13 @@ def _get_export_path(cls, export_path: Optional[pathlib.Path]) -> pathlib.Path:
...
-        if shutil.which("openssl") is None:
-            raise IOError("OpenSSL executable is not present on system")
-
     def _run_openssl_command(self, command: Sequence[Union[str, pathlib.Path]]):
         process = None
         cli_app = self.get_current_cli_app()
         try:
             if cli_app:
-                process = cli_app.execute(command, [command[-1]])
+                obfuscate_patterns = [arg for arg in command if str(arg)[:5] == "pass:"]
```
artemii-yanushevskyi, line 95:
```suggestion
                obfuscate_patterns = [arg for arg in command if str(arg).startswith("pass:")]
```

artemii-yanushevskyi (approved):

src/codemagic/models/certificate_p12_exporter.py
priitlatt: Good catch. I managed to mix up `str.startswith` and [`str.removeprefix`](https://docs.python.org/3.9/library/stdtypes.html#str.removeprefix). The latter cannot be used as it requires Python 3.9, but our lowest bound is 3.7.

PR #415, opened 2024-07-04
Add support for certificate type `DEVELOPER_ID_APPLICATION_G2`

### LINKED ISSUE #414: "app-store-connect fetch-signing-files --type MAC_APP_DIRECT" does not fetch new DEVELOPER_ID_APPLICATION_G2 certificates
`app-store-connect fetch-signing-files --type MAC_APP_DIRECT` calls `GET https://api.appstoreconnect.apple.com/v1/certificates {'limit': 100, 'sort': 'displayName', 'filter[certificateType]': 'DEVELOPER_ID_APPLICATION'}`.  But newer Developer ID Application certificates created by choosing the `G2 Sub-CA` profile type don't match the filter because their type is `DEVELOPER_ID_APPLICATION_G2`.

It looks like it's impossible to use `DEVELOPER_ID_APPLICATION_G2` certificates with Codemagic.

priitlatt:
**Resolves #414** 

When creating Developer ID Application certificate in Apple Developer Portal, it is possible to select "G2 Sub-CA" for its profile type:

Screenshot 2024-07-04 at 12 15 32 (1)

The certificate that is created will have type `DEVELOPER_ID_APPLICATION_G2`, which as of 04.07.24 is not documented as a possible [certificate type](https://developer.apple.com/documentation/appstoreconnectapi/certificatetype), but is used nonetheless.

Changes in this PR add bindings for this type and make it possible to use such certificates when requesting certificates for provisioning profiles of type `MAC_APP_DIRECT` and `MAC_CATALYST_APP_DIRECT`.

**Updated actions:**
- `app-store-connect fetch-signing-files`
- `app-store-connect certificates list`

### DIFF
--- a/src/codemagic/apple/resources/enums.py
+++ b/src/codemagic/apple/resources/enums.py
@@ -1,7 +1,12 @@
 from __future__ import annotations
 
 import re
+from collections import OrderedDict
+from typing import List
+from typing import Optional
+from typing import Sequence
 from typing import Tuple
+from typing import Union
 
 from codemagic.models.enums import ResourceEnum
 from codemagic.utilities.decorators import deprecated
@@ -157,7 +162,13 @@ class CapabilityType(ResourceEnum):
 
 
 class CertificateType(ResourceEnum):
+    """
+    https://developer.apple.com/documentation/appstoreconnectapi/certificatetype
+    """
+
     DEVELOPER_ID_APPLICATION = "DEVELOPER_ID_APPLICATION"
+    # Undocumented developer ID certificate with profile type "G2 Sub-CA"
+    DEVELOPER_ID_APPLICATION_G2 = "DEVELOPER_ID_APPLICATION_G2"
     DEVELOPER_ID_KEXT = "DEVELOPER_ID_KEXT"
     DEVELOPMENT = "DEVELOPMENT"
     DISTRIBUTION = "DISTRIBUTION"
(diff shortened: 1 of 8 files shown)


src/codemagic/tools/app_store_connect/action_groups/certificates_action_group.py
```diff
@@ -166,10 +165,23 @@ def list_certificates(
             raise AppStoreConnectError("Cannot create or save resource without certificate private key")
 
         _certificate_type: Optional[CertificateType] = _deprecated_kwargs.get("certificate_type")
-        certificate_types_filter = self._resolve_certificate_types(_certificate_type, certificate_types, profile_type)
+        if isinstance(_certificate_type, CertificateType):
+            warning = (
+                "Deprecation warning! Keyword argument "
+                '"certificate_type: Optional[CertificateType]" is deprecated in favor of '
+                '"certificate_types: Optional[Union[CertificateType, Sequence[CertificateType]]] = None", '
```
artemii-yanushevskyi, line 172:
```suggestion
                '"certificate_types: Optional[Union[CertificateType, Sequence[CertificateType]]]", '
```

tests/apple/resources/enums/test_certificate_type.py
```diff
@@ -10,3 +13,147 @@ def test_from_profile_type(profile_type: ProfileType):
...
+
+@pytest.mark.parametrize("certificate_type", list(CertificateType))
+def test_resolve_using_literal_certificate_type(certificate_type: CertificateType):
+    """
+    Check that type can be resolved when CertificateType literal instance is passed via
+    `certificate_types` keyword argument. Expected result is to have a list that contains
+    only the passed type.
+    """
+    resolved_types = CertificateType.resolve(certificate_types=certificate_type)
+    assert resolved_types == [certificate_type]
```
artemii-yanushevskyi, line 26: Shouldn't it be
```suggestion
    assert resolved_types == certificate_type
```
instead?

Since `certificate_type` is already a list of all possible certificate types?

src/codemagic/apple/resources/enums.py
```diff
@@ -200,6 +211,39 @@ def from_profile_type(cls, profile_type: ProfileType) -> CertificateType:
         else:
             raise ValueError(f"Certificate type for profile type {profile_type} is unknown")
 
+    @classmethod
+    def resolve(
```
artemii-yanushevskyi, line 215: It's not quite obvious what we resolve. Consider naming it like `list_expected_certificate_types`. Or just add a docstring with a brief explanation...

artemii-yanushevskyi (commented):
Left a few minor remarks

tests/apple/resources/enums/test_certificate_type.py
priitlatt: Nope, we parametrize over the values in `CertificateType` enumeration. Each `certificate_type` is an instance of `CertificateType`. Something like
```python
>>> class E(Enum):
...   A = 'A'
...   B = 'B'
...
>>> list(E)
[<E.A: 'A'>, <E.B: 'B'>]
```

src/codemagic/apple/resources/enums.py
priitlatt: Updated in [eca517b](https://github.com/codemagic-ci-cd/cli-tools/pull/415/commits/eca517be666856c4f1f7b3f50cceb08bdc61ee00).

src/codemagic/tools/app_store_connect/actions/fetch_signing_files_action.py
```diff
@@ -218,7 +207,7 @@ def _create_missing_profiles(
         platform: Optional[BundleIdPlatform] = None,
     ) -> Iterator[Profile]:
         if not bundle_ids_without_profiles:
-            return []
+            return
```
priitlatt, line 210: That was suddenly picked up by `mypy`. This is a generator function and return values are meaningless as the function does not return a single value at any point, but a generator object. This return just ensures that we're never going to reach a `yield` statement.

artemii-yanushevskyi (approved):

PR #421, opened 2024-08-02
Update project-id to project-number

fran-tirapu:
## QA

Previous `project-id` and new `project-number` arguments are used in couple of use cases and behaviour is the same after the name change:

-  If argument is not passed on previous version, --project-id is requested as mandatory:

```
$> firebase-app-distribution get-latest-build-version                                                                                
usage: firebase-app-distribution get-latest-build-version [-h] [--log-stream {stderr,stdout}] [--no-color] [--version] [-s] [-v] --app-id APP_ID
                                                          --project-id PROJECT_ID [--credentials CREDENTIALS]
firebase-app-distribution get-latest-build-version: error: the following arguments are required: --app-id/-a, --project-id/-p

```

```
$> firebase-app-distribution releases list                                                                                        
usage: firebase-app-distribution releases list [-h] [--log-stream {stderr,stdout}] [--no-color] [--version] [-s] [-v] --app-id APP_ID
                                               [--limit LIMIT] [--order-by {createTimeDesc,createTime}] [--json] --project-id PROJECT_ID
                                               [--credentials CREDENTIALS]
firebase-app-distribution releases list: error: the following arguments are required: --app-id/-a, --project-id/-p
```

Now same output is returned but with  `--project-number/-p`

- Several test has been done to obtain same response from Firebase API without success, because you need a valid iOS or Android app published to obtain a valid http response 200. At least we can test that final URL composition is the same and returned http response error code is also the same.

```
$> firebase-app-distribution get-latest-build-version --project-number 123456 --app-id 654321

GET https://firebaseappdistribution.googleapis.com/v1/projects/123456/apps/654321/releases?orderBy=createTimeDesc&pageSize=1&alt=json --> HttpError 400 Request contains an invalid argument.
```

```
$> firebase-app-distribution releases list --project-number 123456 --app-id 654321 

GET https://firebaseappdistribution.googleapis.com/v1/projects/123456/apps/654321/releases?orderBy=createTimeDesc&pageSize=25&alt=json --> HttpError 400 Request contains an invalid argument.
```


artemii-yanushevskyi:
For backwards compatibility, please keep `--project-id` flag, and `project_id` argument in public Python API. If the deprecated flag or the argument is used, a deprecation message should be shown.

priitlatt (changes requested):
Exactly what Artemii said regarding backwards compatibility. It would be OK and even preferred to only show the current configuration option in Markdown docs and terminal help messages, but the old version must remain functional for the time being. Otherwise we're going to break workflows that are working fine as of now.

fran-tirapu:
The job done on #422 give us the new option `mutually_exclusive_group`.
Docs are generated properly and arguments are automatically checked by [argparse](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) standard lib.

priitlatt (approved):
🚀

PR #424, opened 2024-08-07
Fix using xcresulttool on Xcode 16.0 beta 3+

priitlatt:
With Xcode 16.0 beta 3+ part of `xcresulttool` API that we use to parse Xcode test results has been deprecated:
```shell
% xcrun xcresulttool get object --help
OVERVIEW: Get Result Bundle Object. This subcommand is deprecated and will be removed in a future release, consider using `xcresulttool get test-report` instead.

USAGE: xcresulttool get object [--legacy] --path <path> [--id <id>] [--version <version>] [--format <format>]
...
```

This causes `xcresulttool get object` calls with recent Xcode versions to exit with error and consequently some actions in `xcode-project` fail.

Apple has provided a quick workaround by providing `--legacy` flag that can be used to still use the old API even on more recent `xcresulttool` version. Changes on this PR make use of that flag when calling `xcresulttool get object` in case `xcresulttool` version is at least `23021`.

**Updated actions:**
- `xcode-project run-tests`
- `xcode-project test-summary`
- `xcode-project junit-test-results`

**QA notes**

Test are running and results can be successfully parsed using the following Xcode versions:
- ✅ Xcode 15.3 (15E204a)
- ✅ Xcode 15.4 (15F31d)
- ✅ Xcode 16.0 (16A5211f)

### DIFF
--- a/src/codemagic/models/xctests/xcresulttool.py
+++ b/src/codemagic/models/xctests/xcresulttool.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import json
 import pathlib
+import re
 import subprocess
+from functools import lru_cache
 from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING
 from typing import Any
@@ -11,9 +13,12 @@ from typing import List
 from typing import Optional
 from typing import Sequence
 
+from packaging.version import Version
+
 from codemagic.cli import CommandArg
 from codemagic.mixins import RunningCliAppMixin
 from codemagic.mixins import StringConverterMixin
+from codemagic.utilities import log
 
 if TYPE_CHECKING:
     from codemagic.cli import CliApp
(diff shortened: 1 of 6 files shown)


src/codemagic/models/xctests/xcresulttool.py
```diff
@@ -26,6 +31,45 @@ def __init__(self, message: str, stderr: str):
...
 
 class XcResultTool(RunningCliAppMixin, StringConverterMixin):
+    @classmethod
+    @lru_cache(1)
+    def get_tool_version(cls) -> Optional[Version]:
+        # Cache the return value as it is not expected that chosen Xcode version,
+        # and consequently xcresulttool version, is changed during the execution
+        # of the action which requires xcresulttool. As there are possibly quite
+        # a few xcresulttool invocations for which we also need to know the used
+        # version, it is better to do the version check only once.
```
mohammedbabelly20, line 41:
```suggestion
        # Cache xcresulttool version to avoid repeated checks.
        # Assumes Xcode (and thus xcresulttool) version remains constant during execution.
```

src/codemagic/models/xctests/xcresulttool.py
```diff
@@ -26,6 +31,45 @@ def __init__(self, message: str, stderr: str):
...
+            log.get_file_logger(cls).error("Failed to capture xcresulttool version from %r", version_output)
+            return None
+
+        return Version(match.group("version"))
+
+    @classmethod
+    def _requires_legacy_flag(cls) -> bool:
+        """
+        Starting from Xcode 16 beta 3 'xcresulttool get --format json' has been deprecated and
+        cannot be used without '--legacy' flag.
```
mohammedbabelly20, line 65:
```suggestion
        Starting from Xcode 16 beta 3 'xcresulttool get' has been deprecated and
        cannot be used without '--legacy' flag.
```

src/codemagic/models/xctests/xcresulttool.py
mohammedbabelly20: > Long comments are often a code smell, indicating that the code itself may be too complex or unclear.

Which is not the case here, so a short version is suffice.

PR #428, opened 2024-09-20
Fix `keychain add-certificates` on macOS 15

priitlatt:
On macOS 15.0 `security import -f pkcs12` does not accept unencrypted PKCS#12 containers that are generated by OpenSSL. The command terminates with exist code 1 and emits an error:
```shell
% security import certificate.p12 -f pkcs12 -k my-keychain.keychain-db -P "" -T /usr/bin/codesign -T /usr/bin/productsign
security: SecKeychainItemImport: Unable to decode the provided data.
```

The very same certificate can be still imported when `-f` switch value is changed to `openssl`:
```shell
% security import certificate.p12 -f openssl -k my-keychain.keychain-db -P "" -T /usr/bin/codesign -T /usr/bin/productsign
1 key imported.
1 certificate imported.
```

Alter action `keychain add-certificates` to use attempt certificate import with `openssl` format if initial import with `pkcs12` format fails such that error message contains `Unable to decode the provided data`.

**Updated actions:**
- `keychain add-certificates`


mikhail-tokarev (approved):
LGTM

PR #431, opened 2024-10-11
Feature/xcresulttool test reports

priitlatt:


src/codemagic/models/xctests/converter.py
```diff
@@ -140,7 +174,169 @@ def actions_invocation_record_to_junit(cls, actions_invocation_record: ActionsIn
...
+
+    @classmethod
+    def _get_test_node_duration(cls, xc_test_case: XcTestNode) -> float:
+        if not xc_test_case.duration:
+            return 0.0
+
+        duration = xc_test_case.duration.replace(",", ".")
+        if duration.endswith("s"):
+            duration = duration[:-1]
+        return float(duration)
```
fran-tirapu, line 251:
```suggestion
        return float(re.sub(r"[^\d\.,]", "", xc_test_case.duration).replace(",", "."))
```

src/codemagic/models/xctests/converter.py
```diff
@@ -140,7 +174,169 @@ def actions_invocation_record_to_junit(cls, actions_invocation_record: ActionsIn
...
+        xc_test_suite: XcTestNode,
+        xc_test_result_summary: XcSummary,
+    ) -> List[Property]:
+        device = cls._get_run_destination(xc_test_suite)
+
+        properties: List[Property] = [Property(name="title", value=xc_test_suite.name)]
+        if xc_test_result_summary.start_time:
+            Property(name="started_time", value=cls._timestamp(xc_test_result_summary.start_time))
+        if xc_test_result_summary.finish_time:
+            Property(name="ended_time", value=cls._timestamp(xc_test_result_summary.finish_time))
```
fran-tirapu, line 290: `properties.append` seems missed here

tests/models/xctests/converter/test_xcode_16_converter.py
```diff
@@ -0,0 +1,325 @@
...
+from codemagic.models.junit import Skipped
+from codemagic.models.junit import TestCase
+from codemagic.models.junit import TestSuites
+from codemagic.models.xctests import XcResultTool
+from codemagic.models.xctests.converter import Xcode16XcResultConverter
+
+
+def _get_test_report_tests(_):
+    return {
+        "devices": [
```
fran-tirapu, line 18: consider charge those dicts from an external json file

fran-tirapu (changes requested):

src/codemagic/models/junit/definitions.py
```diff
@@ -30,13 +30,17 @@ def disabled(self) -> int:
         return sum(suite.disabled for suite in self.test_suites if suite.disabled)
 
     @property
-    def errors(self) -> int:
+    def errors(self) -> Optional[int]:
```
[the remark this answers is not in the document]
priitlatt: Errors and failures in JUnit XML format are optional. By having optional integer value for those here we can more closely represent the reality. Xcode 16+ does not tell anything about `failed` test cases, there are only `errored` tests. So the corresponding JUnit XML test suite should not include `failures` at all for such cases. But if we'd always return `0` then it is not possible to differentiate the cases where test suite does not specify this value from those where it is defined but legitemately `0`.

Ultimately it is to omit undefined failures or errors from the test result overview tables.

src/codemagic/models/xctests/converter.py
priitlatt: Yes, you're very much correct on that one.

priitlatt: Fixed in [d115c03](https://github.com/codemagic-ci-cd/cli-tools/pull/431/commits/d115c032e3e5cbfeea766bf48a25a80e982527d5).

src/codemagic/models/xctests/converter.py
priitlatt: I'd rather not use this loose regex pattern. It can do transformations that are not intentional and hide values which we're currently not supporting. Current "dumb" approach can handle only strings that contain seconds, but I really have no idea whether Xcode represents longer lasting tests (more than couple of minutes) in some other way. When unsupported string representation comes, then this raises an error and hopefully we'll get to know about it in one way or another.

tests/models/xctests/converter/test_xcode_16_converter.py
priitlatt: Changed in [8ced5b1](https://github.com/codemagic-ci-cd/cli-tools/pull/431/commits/8ced5b1e6ff6ed3cb6f438ad43c94e97d2cceec5).

fran-tirapu (approved):
LGTM

I just suggest to add a comment on CHANGELOG about statistics treatment  as new Xcode versions does not use "failure" status for tests.

PR #443, opened 2025-02-13
Unify Google API clients for Google Play and Firebase

priitlatt:
Unify Google APIs usage under a common package `codemagic.google` which can be used for both `firebase-app-distribution` and `google-play` tools.

> [!NOTE]  
> This PR contains some **breaking** changes to internal Python APIs, but all **CLI usage remains backwards compatible**.

**Updated actions:**
- `firebase-app-distribution get-latest-build-version`
- `firebase-app-distribution releases list`
- `google-play get-latest-build-number`
- `google-play tracks get`
- `google-play tracks list`
- `google-play tracks promote-release`

For more detailed list of changes refer to `CHANGELOG.md`.

**QA**

Full feature parity with current production version is ensured on updated actions.


src/codemagic/apple/resources/resource.py
```diff
@@ -34,34 +34,12 @@ class ResourceId(str):
     pass
 
 
-class DictSerializable:
-    _OMIT_KEYS: Tuple[str, ...] = ("_raw",)
-    _OMIT_IF_NONE_KEYS: Tuple[str, ...] = tuple()
-
+class AppleDictSerializable(DictSerializable):
```
priitlatt, line 37: `DictSerializable` is now defined in `codemagic.models`. It is identical to what is here apart from how `datetime` objects are serialized.

src/codemagic/google/google_play_client.py
priitlatt, line 1: Replaces `GooglePlayDeveloperAPIClient` from `codemagic.google_play`.

src/codemagic/google/resource_managers/firebase/firebase_releases_manager.py
priitlatt, line 1: Functionally identical to old `ReleaseManager` that was defined in `src/codemagic/google/resource_managers/release_manager.py`.

src/codemagic/google/resource_managers/__init__.py
priitlatt, line 1: Package `codemagic.google.resource_managers` was refactored to `codemagic.google.services` to better reflect how Google organizes their APIs.

src/codemagic/google/resources/identifiers/app_identifier.py
priitlatt, line 1: This isn't needed at all. Service can build the parent URI from explicit project number and app ID arguments without another layer of abstraction.

src/codemagic/google/services/firebase/releases_service.py
priitlatt, line 1: That used to be `ReleasesManager` before.

src/codemagic/google/services/resource_service.py
priitlatt, line 1: Method `_execute_request` from here was moved to new base service class `ResourceService`.

src/codemagic/google/resources/resource.py
```diff
@@ -1,59 +1,61 @@
...
 
-class OrderBy(ResourceEnum):
-    CREATE_TIME_DESC = "createTimeDesc"
-    CREATE_TIME_ASC = "createTime"
-
-
-class Resource(ABC):
-    __google_api_label__: ClassVar[Optional[str]] = None
+class Resource(DictSerializable, JsonSerializable):
+    @staticmethod
```
fran-tirapu, line 13: is there any other purpose than the `str` representation of the object for all these code?
Now the object is `DictSerializable/JsonSerializable` so probably we can get rid of all this and just use a prettify version of the `dict/json`.

fran-tirapu (approved):
Good refactoring job and nice comments to help the reviewing job.
I left couple of comments.

src/codemagic/google/resources/resource.py
priitlatt: Base `Resource` class is indeed mostly just to have a nice string representation of the object which is displayed to the user.

<details>
<summary>Screenshot</summary>

Screenshot 2025-02-20 at 15 00 03

</details>

`JsonSerializable` is just nice to have and adds a convenience method `resource.json(**dump_options)`, which is just a shorthand for `json.dumps(resource.dict(), **dump_options)`. As it exists and is also widely used for App Store Connect resources, then I don't see any reasons not to have it here.

`DictSerializable` does a little more than just `dataclasses.asdict` though. It generates a dictionary representation of the resource in a way that the dictionary can be directly passed along to different API calls which require the dictionary to be JSON serializable. That is, the whole dictionary is recursively transformed to contain only basic JSON serializable types. For example ISO8601 representation is used for datetimes, enums are converted to their respective values etc.

PR #446, opened 2025-02-21
Feature: Add actions to publish APKs and Android App Bundles to Google Play

priitlatt:
**Add actions to publish APKs and Android App Bundles to Google Play.**

Highlights of this PR are the `google-play apks publish`, `google-play bundles publish` and `google-play internal-app-sharing upload` actions that allow users to publish APKs and App Bundles either to Google Play release tracks or share internally to their team via Google Play Internal App Sharing.

**Example usages:**

<details>
<summary>Internal App Sharing</summary>

```shell
% google-play internal-app-sharing upload --apk $APK
Upload APK "/tmp/cli_tools_google_play_android/app/release/app-release.apk" to Google Play internal app sharing
App name: CLItools Google Play test app
Certificate issuer: None
Certificate subject: None
Debuggable: No
Min OS version: 24
Package name: io.codemagic.cli_tools_google_play
Version code: 8
Version: 1.0

Uploaded APK to Google Play internal app sharing
-- InternalAppSharingArtifact --
Download url: https://play.google.com/apps/test/RQ0leA8LLuQ/ahAO29uNTGlMyxMuX-LgdW05BNSmstYGxxnGTs1tJPIxCbwXk_1XSFIGXzIfCSsAVt3DoQQS-NDFbGstZSIml5XxNC
Certificate fingerprint: SHA256: ED:B6:58:E3:53:A7:4C:30:F8:5C:11:16:E3:53:36:C2:98:75:50:E6:08:C1:EE:EB:ED:7C:91:45:E4:99:23:E3
Sha256: be69Vn8JTZt-e_lKKcemNkd211pzCkapuSE3oSIudOY
```
</details>

<details>
<summary>Publish Android App Bundle</summary>

```shell
% google-play bundles publish \
    --bundle $BUNDLE \
    --track "alpha" 
    --release-name "Release #13" 
    --release-notes "@file:google_play_release_notes.json" \
    --rollout-fraction 0.1 \
    --in-app-update-priority 3
Publishing App Bundle "/tmp/cli_tools_google_play_android/app/release/app-release.aab" to Google Play track alpha

Upload App Bundle "/tmp/cli_tools_google_play_android/app/release/app-release.aab" to Google Play
App name: CLItools Google Play test app
Certificate issuer: Organization: Codemagic
Certificate subject: Organization: Codemagic
Debuggable: No
Min OS version: 24
Package name: io.codemagic.cli_tools_google_play
Version code: 13
Version: 1.0

[...]

PR #447, opened 2025-02-27
Fix `android-app-bundle build-apks` and  `android-app-bundle build-universal-apk` actions

priitlatt:
Fix a regression from #446 that caused `android-app-bundle build-apks` and  `android-app-bundle build-universal-apk` actions to fail unexpectedly if keystore arguments were provided.

Keystore and key password arguments were passed to bundletool using invalid flags (originally in version 0.56.0 like [this](https://github.com/codemagic-ci-cd/cli-tools/blob/v0.56.0/src/codemagic/tools/android_app_bundle.py#L477-L484) and in 0.57.0 as [this](https://github.com/codemagic-ci-cd/cli-tools/blob/v0.57.0/src/codemagic/shell_tools/bundletool.py#L129-L138)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-Version 0.57.0
+Version 0.57.1
 -------------
 
 **Bugfixes**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Version 0.57.0
 -------------
 
+**Bugfixes**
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Version 0.57.0
```
priitlatt, line 1:
```suggestion
Version 0.57.1
```

PR #453, opened 2025-03-27
Fix version info resolving when submitting app to App Store

priitlatt:
Submitting macOS applications to App Store review is broken in case the submission is invoked from Python API by specifying platform info using `AppStoreVersionInfo` instead of the `platform` argument. That is, running
```python
AppStoreConnect(...).submit_to_app_store(
    "build-id",
    app_store_version_info=AppStoreVersionInfo(
        platform=Platform.MAC_OS,
        ...,
    ),
    ...,  # platform uses default value
)
```
will end up using default platform type `Platform.IOS` for the review submission instead of the one that is given in version info. This causes a conflict that fails the App Store submission.

The issue is resolved by altering private `_submit_to_app_store` method to only use platform from the `app_store_version_info` where correct platform type is given.

**Updated actions:**
- `app-store-connect submit-to-app-store`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@ Version 0.57.4
 -------------
 
 **Bugfixes**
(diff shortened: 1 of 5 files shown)


fran-tirapu (approved):
🚀

PR #455, opened 2025-04-04
Androguard improvements

priitlatt:
There are couple of problems with [androguard](https://github.com/androguard/androguard) dependency and its usage.

First, updating `androguard` log level multiple times causes `ValueError`s because `androguard.util.set_log` method uses hardcoded ID value to remove `logger`'s handler, but new handler is added gets another ID. 

Changes here ensure that log level is updated only once.

<details>
<summary>Stacktrace</summary>

```python
  File "/home/builder/builder/.venv/lib/python3.10/site-packages/codemagic/models/application_package/apk_package.py", line 63, in _validate_package
    _ = self._apk
  File "/home/builder/.local/share/uv/python/cpython-3.10.16-linux-x86_64-gnu/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/home/builder/builder/.venv/lib/python3.10/site-packages/codemagic/models/application_package/apk_package.py", line 73, in _apk
    return _get_androguard_apk(self.path)
  File "/home/builder/builder/.venv/lib/python3.10/site-packages/codemagic/models/application_package/apk_package.py", line 56, in _get_androguard_apk
    set_log("ERROR")
  File "/home/builder/builder/.venv/lib/python3.10/site-packages/androguard/util.py", line 37, in set_log
    logger.remove(0)
  File "/home/builder/builder/.venv/lib/python3.10/site-packages/loguru/_logger.py", line 1050, in remove
    raise ValueError("There is no existing handler with id %d" % handler_id) from None
ValueError: There is no existing handler with id 0
```
</details>

Secondly, currently there is no way to control `androguard` availability ahead of time and consequently `google-play` actions that handle APK files can behave unexpectedly when they first need to install the dependency.

To overcome that, add new action `codemagic-cli-tools ensure-androguard` which installs `androguard` if it isn't present in CLI tools environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,17 @@
 Version 0.57.6
 -------------
 
+This release contains changes from [PR #455](https://github.com/codemagic-ci-cd/cli-tools/pull/455)
+
+**Features**
+- Add action `codemagic-cli-tools ensure-androguard` to pre-install [`androguard`](https://github.com/androguard/androguard) dependency that is required for working with APK files.
+
 **Bugfixes**
-- Fix `ValueError` when initializing more than one `ApkPackage` instances using Python APIs. [PR #454](https://github.com/codemagic-ci-cd/cli-tools/pull/454)
+- Fix `ValueError` when initializing more than one `ApkPackage` instances using Python APIs.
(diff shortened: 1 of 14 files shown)


mikhail-tokarev (approved):
LGTM

PR #457, opened 2025-04-09
Fix certificate issuer and subject formatting

priitlatt:
If APK or Android App Bundle contains a certificate which includes unknown attributes, then showing the summary fails package summary fails:

```python
  ...
  File "codemagic/models/application_package/apk_package.py", line 120, in get_summary
  File "codemagic/models/application_package/abstract_package.py", line 60, in _format_name
  File "codemagic/models/application_package/abstract_package.py", line 59, in <genexpr>
KeyError: '<unexpected.attribute>'
```

Use only known RFC 4514 certificate attributes to construct the human readable representations or issuer and subject.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Version 0.57.7
 -------------
 
 **Bugfixes**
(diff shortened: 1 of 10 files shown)


mikhail-tokarev (approved):
🚀

PR #465, opened 2025-05-23
Fix parsing datetime with nanoseconds in Apple API response

remarkov:
Apparently Apple started to use nanoseconds in the response payload which cannot be parsed properly with stdlib without string manipulations, thus producing the errors like
```
ValueError time data '2025-05-23T09:50:03.168550349-07:00' does not match format '%Y-%m-%dT%H:%M:%S.%f%z'
```

As there is already some inconsistency in the API responses that requires additional logic, instead of complicating it further we can introduce a dependency of [dateutil](https://github.com/dateutil/dateutil/) which handles all these properly:

```
In [2]: from dateutil.parser import parse
In [3]: parse("2021-01-28T06:01:32-08:00")
Out[3]: datetime.datetime(2021, 1, 28, 6, 1, 32, tzinfo=tzoffset(None, -28800))
In [4]: parse("2020-08-04T11:44:12.000+0000")
Out[4]: datetime.datetime(2020, 8, 4, 11, 44, 12, tzinfo=tzutc())
In [5]: parse("2025-05-23T09:50:03.168550349-07:00")
Out[5]: datetime.datetime(2025, 5, 23, 9, 50, 3, 168550, tzinfo=tzoffset(None, -25200))
```


priitlatt:
Perhaps add couple of additional test cases too for new cases? Existing tests for datetime conversions can be found here: https://github.com/codemagic-ci-cd/cli-tools/blob/8612665c7b1dd071527b8b0f8517ba0780602daf/tests/apple/resources/test_resource.py#L14-L39

Issue #500, opened 2026-07-14
`firebase-app-distribution get-latest-build-version` fails: API now rejects `orderBy=createTimeDesc`


MehmetKaranlik:
Since ~2026-07-14, `firebase-app-distribution get-latest-build-version` fails with `Request contains an invalid argument.` for all of our apps (Android and iOS app ids alike). It worked the previous day with no changes on our side.

## Root cause

`ReleasesService.list` sends `orderBy=createTimeDesc` (from `OrderBy.CREATE_TIME_DESC` in `src/codemagic/google/resources/firebase/order_by.py`) to `GET https://firebaseappdistribution.googleapis.com/v1/{parent}/releases`.

That value was never documented — the [releases.list reference](https://firebase.google.com/docs/reference/app-distribution/rest/v1/projects.apps.releases/list) specifies a space-separated `desc` suffix, e.g. `createTime desc` — but the backend used to tolerate the camelCase form. The API discovery document's `orderBy` description now also lists `update_time`/`expire_time` as supported order fields, so Google appears to have rolled out a reworked releases API (same rollout family as the unannounced change in #472) whose parser strictly validates `orderBy` and returns `400 INVALID_ARGUMENT` for `createTimeDesc`.

## Example

```
$ firebase-app-distribution get-latest-build-version -p <project-number> -a <app-id>
Request contains an invalid argument.
```

(exit code 1; same for every app in the project)

## Suggested fix

Change the enum values to the documented grammar, e.g.

```python
class OrderBy(ResourceEnum):
    CREATE_TIME_DESC = "createTime desc"
    CREATE_TIME_ASC = "createTime"
```

or simply omit `orderBy` for the default case — `createTime` descending is the API's documented default ordering.

## Workaround

Until this is released, we sed-patch the installed package in our CI step before invoking the tool:

```bash
CM_PY="$(sed -n '1s/^#!//p' "$(command -v firebase-app-distribution)")"
ORDER_BY_FILE="$("$CM_PY" -c 'import codemagic.google.resources.firebase.order_by as m; print(m.__file__)')"
sed -i.bak 's/createTimeDesc/createTime desc/' "$ORDER_BY_FILE"
```

Related: #472 (previous unannounced server-side change to the same endpoint).

priitlatt:
@MehmetKaranlik, thanks for reporting this — the root-cause analysis is spot on.                                                                                                         

[...]

PR #501, opened 2026-07-15
Fix `firebase-app-distribution` release ordering rejected by the Firebase API

priitlatt:
Since around 2026-07-14, `firebase-app-distribution get-latest-build-version` and `firebase-app-distribution releases list` started failing with `Request contains an invalid argument.` for every app, with no changes on our side (reported in #500).

The Firebase App Distribution [`releases.list`](https://firebase.google.com/docs/reference/app-distribution/rest/v1/projects.apps.releases/list) API now rejects the previously tolerated camelCase `createTimeDesc`. The documented grammar uses a space-separated `desc` suffix, i.e. `createTime desc`.

The catch is that `OrderBy.value` was doing double duty: it was both the value sent to the API *and* the token accepted by the public `--order-by` CLI argument. Simply changing it to `createTime desc` would fix the API call but break the CLI (`--order-by createTimeDesc` would stop working and the value would need quoting). This PR decouples the two.

**Changes**

- `codemagic.google.resources.firebase.OrderBy` now holds the documented API grammar (`createTime desc` / `createTime`) — this is the value sent to `releases.list`.
- Introduce a CLI-facing `ReleasesOrderByArgument` in the tool layer (`argument_types.py`) that keeps the single-token values (`createTimeDesc` / `createTime`) and maps to `OrderBy` via its `order_by` property. The `--order-by` argument and the `list_releases` action signatures now use this type, converting to `OrderBy` at the point where the action group calls into the service.

**Effect**

- `firebase-app-distribution get-latest-build-version` and `firebase-app-distribution releases list` work again against the updated API.
- The `--order-by` CLI interface is unchanged and fully backwards compatible: same `createTimeDesc` / `createTime` choices and `createTimeDesc` default, and the generated docs are unchanged.

<details>
<summary>QA reference</summary>

```shell
# CLI surface unchanged — same choices and default token
$ uv run firebase-app-distribution releases list --help | grep -A1 'order-by'     
usage: firebase-app-distribution releases list [-h] (--project-id PROJECT_ID | --project-number PROJECT_NUMBER) [--credentials CREDENTIALS] [--json] --app-id APP_ID [--limit LIMIT] [--order-by {createTimeDesc,createTime}] [--log-stream {stderr,stdout}] [--no-color] [--version] [-s] [-v]                                                                                                                                                                                                   

--
  --order-by, -o {createTimeDesc,createTime}
                        Sort resources in the specified order [Default: createTimeDesc]

# CLI token maps to the corrected wire value
$ uv run firebase-app-distribution get-latest-build-version \
  --app-id "1:146661841143:android:a8d456e0c8b5e71bd11bf2" \
  --project-number "146661841143"
No releases available for 1:146661841143:android:a8d456e0c8b5e71bd11bf2
```
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Version 0.69.0
+-------------
+
+**Bugfixes**
+- Fix `firebase-app-distribution` actions `get-latest-build-version` and `releases list`. The Firebase App Distribution API changed `orderBy` query parameter validation and rejects the previously tolerated `createTimeDesc`, requiring the documented space-separated grammar `createTime desc` instead.
+
 Version 0.68.0
 -------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "codemagic-cli-tools"
-version = "0.68.0"
+version = "0.69.0"
 description = "CLI tools used in Codemagic builds"
 authors = [{ name = "Priit Lätt", email = "priit@nevercode.io" }]
 requires-python = ">=3.8,<4"
--- a/src/codemagic/__version__.py
+++ b/src/codemagic/__version__.py
@@ -1,5 +1,5 @@
 __title__ = "codemagic-cli-tools"
 __description__ = "CLI tools used in Codemagic builds"
-__version__ = "0.68.0.dev"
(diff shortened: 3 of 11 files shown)


mohammedbabelly20 (approved):
LGTM

